PR #58, opened 2024-08-19
Catan unittests

PietroPasotti:
Unittests with catan


simskij:
@sed-i, @dstathis: are we ok with taking this in? unless im missing something, it does introduce another unit testing facet that we've not really discussed as a group.

simskij (changes requested):
Folder is named `katan`, while the tool is named `catan`.

PR #70, opened 2024-09-04
Holistic worker logic

PietroPasotti:
This PR largely refactors the worker hook logic to turn it into a more holistic approach.
All logic that is not conditional on a specific event being run is put in a single method called on `Worker.__init__`.
All the observe() calls that are now redundant have been removed.

This should fix some issues we've been seeing in CI where the pod churns at unexpected points in time, and allows us to resolve the issue without resorting to defer().

An alternative to side-effecting in `__init__` would be to put an "observe all events" hook that queues the holistic reconciler. Danger of that approach is that we have less control about when exactly it gets run relative to other event handlers, which in turn might be relying on its side-effects

### DIFF
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -171,8 +171,8 @@ class Worker(ops.Object):
             if self._resources_requests_getter
             else None
         )
-        # holistic update logic
-        self._holistic_update()
+        # holistic update logic, aka common exit hook
+        self._reconcile()
 
         # Event listeners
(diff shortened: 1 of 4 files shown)


src/cosl/coordinated_workers/worker.py
```diff
@@ -170,14 +170,11 @@ def __init__(
             if self._resources_requests_getter
             else None
         )
+        # holistic update logic
+        self._holistic_update()
```
sed-i, line 174: "common exit hook" is fun too, but the established term for this is `reconcile`. Correct me if I'm wrong @ca-scribner?

PR #74, opened 2024-09-06
Prevent unnecessary restarts

michaeldmitry:
## Issue
On every event, the workload container gets restarted due to `_update_tls_certificates` and `_set_pebble_layer` always returning True to restart.
- `_update_tls_certificates`
the logic almost always returns True even though the data in relation data might not be different than in the disk
- `_set_pebble_layer`
The `checks` layer is almost always different since pebble automatically adds a `threshold: 3` to the current layer and we don't add that to the new layer.

## Solution
- Add checks in `_update_tls_certificates` to compare relation data with data on disk
- Add `threshold: 3` to `_add_readiness_check`

### DIFF
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -371,6 +371,7 @@ class Worker(ops.Object):
 
         current_plan = self._container.get_plan()
         new_layer = self._pebble_layer()
+
         self._add_readiness_check(new_layer)
 
         def diff(layer: Layer, plan: Plan):
@@ -394,7 +395,13 @@ class Worker(ops.Object):
             return
 
         new_layer.checks["ready"] = Check(
-            "ready", {"override": "replace", "http": {"url": self._readiness_check_endpoint(self)}}
+            "ready",
+            {
+                "override": "replace",
+                # threshold gets added automatically by pebble
+                "threshold": 3,
+                "http": {"url": self._readiness_check_endpoint(self)},
+            },
         )
 
     def _reconcile(self):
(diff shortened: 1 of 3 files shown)


src/cosl/coordinated_workers/worker.py
```diff
@@ -394,7 +395,13 @@ def _add_readiness_check(self, new_layer: Layer):
             return
 
         new_layer.checks["ready"] = Check(
-            "ready", {"override": "replace", "http": {"url": self._readiness_check_endpoint(self)}}
+            "ready",
+            {
+                "override": "replace",
+                # threshold gets added automatically by pebble
+                "threshold": 3,
```
PietroPasotti, line 402: I was convinced 3 was the default value, but either way, good to make it explicit

tests/test_coordinated_workers/test_worker.py
```diff
@@ -408,3 +420,90 @@ def __init__(self, framework: Framework):
...
+    )
+    # WHEN the charm receives any event and there are no changes to the config or the layer,
+    #  but some of the services are down
+    container = Container(
+        "foo",
+        can_connect=True,
+    )
+    ctx.run("update_status", State(containers=[container], relations=[relation]))
+
+    assert restart_mock.call_count == 0
```
PietroPasotti, line 453:
```suggestion
    assert not restart_mock.called
```

PietroPasotti (approved):
lgtm, couple nits

src/cosl/coordinated_workers/worker.py
michaeldmitry: yes, but we compare dicts, so we should be explicit

PR #75, opened 2024-09-10
Add safeguard when resources patch is not ready

michaeldmitry:
## Issue
Intermittent issues in CI where the worker goes into error state during `start` hook.
There could be a race between the resource patch and pebble operations i.e., charm code proceeds beyond a can_connect guard, and then lightkube patches the statefulset and the workload is no longer available.

## Solution
Add a safeguard to not perform pebble operations if the patch is not done.

### DIFF
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -405,12 +405,13 @@ class Worker(ops.Object):
         )
 
     def _reconcile(self):
-        """Run all unconditional logic."""
+        """Run all logic that is independent of what event we're processing."""
         # There could be a race between the resource patch and pebble operations
         # i.e., charm code proceeds beyond a can_connect guard, and then lightkube patches the statefulset
         # and the workload is no longer available
+        # `resources_patch` might be `None` when no resources requests or limits are requested by the charm.
         if self.resources_patch and not self.resources_patch.is_ready():
-            logger.debug("Resource patching is not yet ready. Skipping config update.")
+            logger.debug("Resource patch not ready yet. Skipping reconciliation step.")
             return
         self._update_cluster_relation()
         self._update_config()
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -406,6 +406,12 @@ class Worker(ops.Object):
 
     def _reconcile(self):
         """Run all unconditional logic."""
+        # There could be a race between the resource patch and pebble operations
+        # i.e., charm code proceeds beyond a can_connect guard, and then lightkube patches the statefulset
+        # and the workload is no longer available
+        if self.resources_patch and not self.resources_patch.is_ready():
+            logger.debug("Resource patching is not yet ready. Skipping config update.")
+            return
         self._update_cluster_relation()
         self._update_config()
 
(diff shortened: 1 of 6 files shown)


src/cosl/coordinated_workers/worker.py
```diff
@@ -406,6 +406,12 @@ def _add_readiness_check(self, new_layer: Layer):
 
     def _reconcile(self):
         """Run all unconditional logic."""
+        # There could be a race between the resource patch and pebble operations
+        # i.e., charm code proceeds beyond a can_connect guard, and then lightkube patches the statefulset
+        # and the workload is no longer available
+        if self.resources_patch and not self.resources_patch.is_ready():
```
PietroPasotti, line 412:
```suggestion
        if not self.resources_patch.is_ready():
```
or when could the `resources_patch` variable be falsy? please add a comment to explain

src/cosl/coordinated_workers/worker.py
```diff
@@ -406,6 +406,12 @@ def _add_readiness_check(self, new_layer: Layer):
...
+            logger.debug("Resource patching is not yet ready. Skipping config update.")
```
PietroPasotti, line 413:
```suggestion
            logger.debug("Resource patch not ready yet. Skipping reconciliation step since the pod will shortly be restarted.")
```

PietroPasotti: I also think we should surface this in the status. something like "waiting for resource patch to apply". other charms do that too I think

src/cosl/coordinated_workers/worker.py
```diff
@@ -406,6 +406,12 @@ def _add_readiness_check(self, new_layer: Layer):
 
     def _reconcile(self):
         """Run all unconditional logic."""
```
PietroPasotti, line 408: funny that this says - run all unconditional logic - and the first thing we see two lines below is... a conditional! perhaps rephrase it to say something a bit more precise like "logic that does not depend on what event we're processing"

PietroPasotti (changes requested):
some minor nits, the only thing I really think we should change with this PR is surface this information better to the user by means of a (Waiting) status message.

really awesome that you found the root cause of the issue! I had it in CI once but couldn't repro

src/cosl/coordinated_workers/worker.py
michaeldmitry: So, that should already happen. `line 277 inside _on_collect_status()`, The status would be WaitingStatus("waiting 
 for bla bla") or BlockedStatus("failed bla bla").

src/cosl/coordinated_workers/worker.py
michaeldmitry: `resources_patch` can be None when no resources are passed to the worker (don't set requests or limits). Added a comment

src/cosl/coordinated_workers/worker.py
michaeldmitry: I would change the debug message a bit and not mention "since the pod will shortly be restarted.", as it might have failed. that would be captured in the status message

src/cosl/coordinated_workers/worker.py
michaeldmitry: updated docstring

PietroPasotti (approved):

PR #77, opened 2024-09-12
Refactor status logic

PietroPasotti:
- worker will now give priority to 'specific' statuses e..g. waiting on container, blocked (missing relation)... and if none of those are applicable, fall back to an atomic 'check for readiness' 
- worker will now block right after restarting, waiting for the worker process to report ready using the readiness check if provided.

### DIFF
--- a/.github/workflows/distributed-solutions-tests.yaml
+++ b/.github/workflows/distributed-solutions-tests.yaml
@@ -1,0 +1,73 @@
+name: Distributed Solutions Tests
+
+on:
+  push:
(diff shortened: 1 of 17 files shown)


lucabello (approved):

PietroPasotti:
worker failure is expected, as the scenario tests also verify the statuses, which are changing with this PR. shame that we can't run the integration tests because of those. Perhaps we should get rid of the status tests since that behaviour is now tested in the worker.

PR #81, opened 2024-09-18
fix: use https in worker readiness when tls is enabled

lucabello:
## Issue
Currently, the readiness check fails when an HA solution is connected to TLS. This happens because the URL used for the readiness check is provided by the individual worker charm in the `readiness_check_endpoint` constructor argument.

However, the worker charm doesn't know if it's using TLS before instantiating the Worker object (because that's the object that handles TLS).

## Solution

The worker shared object must replace the scheme to use `https` when TLS is enabled, before executing the readiness check.

## Upgrade Notes

This won't break the default `http` behavior, but to make sure this is fully working for `https`, all the worker charms should probably pass `socket.getfqdn()` instead of `localhost` in the `readiness_check_endpoint`.

### DIFF
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -13,6 +13,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict, Union
 from urllib.error import HTTPError
+from urllib.parse import urlparse
 
 import ops
 import tenacity
@@ -288,7 +289,12 @@ class Worker(ops.Object):
             )
 
         try:
-            with urllib.request.urlopen(check_endpoint(self)) as response:
+            # Use "https" if tls is enabled
+            check_url = check_endpoint(self)
+            if self.is_tls_enabled:
+                # _replace is part of the public API, it's not a private method
+                check_url = urlparse(check_url)._replace(scheme="https").geturl()
+            with urllib.request.urlopen(check_url) as response:
                 html: bytes = response.read()
 
             # ready response should simply be a string:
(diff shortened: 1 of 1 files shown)


src/cosl/coordinated_workers/worker.py
```diff
@@ -316,6 +322,11 @@ def check_readiness(self) -> ServiceEndpointStatus:
             logger.exception("Unexpected exception getting readiness endpoint")
         return ServiceEndpointStatus.down
 
+    @property
+    def is_tls_enabled(self) -> bool:
+        """Check if the worker has the TLS certificate on disk."""
+        return self._container.exists(CERT_FILE)
```
PietroPasotti, line 328: I think this check is a bit too naive. What if we're handling a certificates-relation-broken? what if there was an issue with deleting the cert and it's still there but invalid? 
compare this with [traefik's method](https://github.com/canonical/traefik-k8s-operator/blob/main/src/charm.py#L367) and [tempo-k8s](https://github.com/canonical/tempo-k8s-operator/blob/main/src/charm.py#L177). why the difference? which one is correct?

(btw, we really should consider generalizing this and/or agreeing on what is **the way** to check for tls readiness)

src/cosl/coordinated_workers/worker.py
```diff
@@ -316,6 +322,11 @@ def check_readiness(self) -> ServiceEndpointStatus:
             logger.exception("Unexpected exception getting readiness endpoint")
         return ServiceEndpointStatus.down
 
+    @property
+    def is_tls_enabled(self) -> bool:
```
PietroPasotti, line 326: I think this should be private

PietroPasotti (changes requested):
LGTM, but I really want a test for this feature, and some documentation for this change (in the docstrings at least).

PietroPasotti:
one more note: [this is how we fixed it in tempo-worker](https://github.com/canonical/tempo-worker-k8s-operator/blob/main/src/charm.py#L96) 
let's double-check (with a test) that this change is compatible with what we are doing (or else enforce that the parent charm is passing the Worker an http url).

src/cosl/coordinated_workers/worker.py
lucabello: Tempo and traefik both have acesss to Cert Handler, the worker does not; we can't do the check the same way; the only other way is to check if the coordinator has put the cert in relation data

lucabello:
Oh so that's why it's a callable! Okay it makes more sense now; I think what you did with Tempo worker is more reliable than this, so I'll implement that in Mimir / Loki as well :)

PR #83, opened 2024-09-20
fix: use https for scrape jobs when tls is available

### LINKED ISSUE #82: TLS breaks self-monitoring
https://github.com/canonical/cos-lib/blob/fb5111a3c7c38937fb1ce2799fb2cb32949886f6/src/cosl/coordinated_workers/coordinator.py#L513

Currently, we use the worker FQDN in the scrape job, without checking for TLS (so we always use `http`). Since we're in the coordinator object, we can check `self.cert_handler` to see whether TLS is enabled, and if so we should replace the scheme with `https`.

lucabello:
## Issue
Closes #82.

## Solution
As the title specifies, we should use the appropriate URL scheme in scrape jobs, i.e., we should `https` when TLS is enabled and `http` otherwise.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "cosl"
-version = "0.0.34"
+version = "0.0.35"
 authors = [
   { name="sed-i", email="82407168+sed-i@users.noreply.github.com" },
 ]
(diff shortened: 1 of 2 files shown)


sed-i (approved):

PietroPasotti (approved):
I don't want to become the "where are the tests?" parrot, but...

lucabello:
Merging this as we're very close to partner cloud testing and it's a small fix rather than a new addition, but you have a point, and I'll try to add a follow-up PR soon!

PR #84, opened 2024-09-26
S3 tls cert

PietroPasotti:
## Issue
The s3 interface may pass us a tls certificate if the connection to the bucket is encrypted.
In my understanding we have no guarantee that that is the same cert we'd be getting by relating to the authority directly. So it may be that we need one certificate for tempo itself, and a different one for encrypting communications with s3.

## Solution
get the cert from the s3 relation and put it in the s3 configuration block as per (for example) [tempo docs](https://grafana.com/docs/tempo/latest/configuration/#storage) 

TBD:
- is this config uniform for loki and mimir?
- we get a CA chain from the s3 charm, but we need a single cert to configure the storage in the coordinator. which cert do we use?

### DIFF
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -28,6 +28,7 @@ from urllib.parse import urlparse
 
 import ops
 import yaml
+from ops import Container
 
 import cosl
 from cosl.coordinated_workers.interface import ClusterProvider, RemoteWriteEndpoint
@@ -64,7 +65,7 @@ from lightkube.models.core_v1 import ResourceRequirements
 
 logger = logging.getLogger(__name__)
 
-S3_CA_CERT_PATH = "./s3_ca.crt"
+S3_TLS_CA_CHAIN_PATH = "/etc/worker/s3_ca.crt"
 
 # The paths of the base rules to be rendered in CONSOLIDATED_ALERT_RULES_PATH
 NGINX_ORIGINAL_ALERT_RULES_PATH = "./src/prometheus_alert_rules/nginx"
@@ -235,6 +236,7 @@ class Coordinator(ops.Object):
             partial(resources_requests, self) if resources_requests is not None else None
         )
         self._container_name = container_name
+        self._container: Container = charm.unit.get_container(container_name)
         self._resources_limit_options = resources_limit_options or {}
         self.remote_write_endpoints_getter = remote_write_endpoints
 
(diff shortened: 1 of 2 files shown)


src/cosl/coordinated_workers/coordinator.py
```diff
@@ -452,6 +455,11 @@ def _s3_config(self) -> Dict[str, Any]:
         s3_config["access_key_id"] = s3_data.pop("access-key")
         s3_config["secret_access_key"] = s3_data.pop("secret-key")
         s3_config["bucket_name"] = s3_data.pop("bucket")
+        ca_chain = s3_data.get("tls-ca-chain")
+        if ca_chain:
+            # put the cacert to disk
+            Path(S3_CA_CERT_PATH).write_text(ca_chain[0])  # TODO: is any cert in the chain good?
```
michaeldmitry, line 461: I dont think it's returned as a list. Its just a concatenated string, so we can dump all in one file and we'll be fine.

src/cosl/coordinated_workers/coordinator.py
```diff
@@ -452,6 +455,11 @@ def _s3_config(self) -> Dict[str, Any]:
...
+            s3_config["tls_ca_path"] = S3_CA_CERT_PATH
```
michaeldmitry, line 462: we should add that to Tempo's config model https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/tempo_config.py#L259

src/cosl/coordinated_workers/coordinator.py
michaeldmitry, line 461: Also, we should push it to the workload container not the charm

src/cosl/coordinated_workers/coordinator.py
```diff
@@ -63,6 +64,8 @@
 
 logger = logging.getLogger(__name__)
 
+S3_CA_CERT_PATH = "./s3_ca.crt"
```
michaeldmitry, line 67: maybe its better to keep the path uniform with other certs? (e.g: /etc/worker/s3_ca.crt)

src/cosl/coordinated_workers/coordinator.py
PietroPasotti: it is a list, I checked. it gets json-loaded by the lib

PietroPasotti: :facepalm:

src/cosl/coordinated_workers/coordinator.py
PietroPasotti: yeah I was still thinking in terms of charm container

src/cosl/coordinated_workers/coordinator.py
michaeldmitry: Right.
Well in that case, picking the 1st cert can be problematic, as it can contain the server cert not even a CA
Can we maybe join all with a `\n\n` like [this](https://github.com/canonical/observability-libs/blob/main/lib/charms/observability_libs/v1/cert_handler.py#L598)?

PietroPasotti: okay, I guess we'll need to try this out somehow

PietroPasotti:
allright this is not the right way. will open a new pr with a fix

PR #86, opened 2024-09-27
Return sorted peers set

michaeldmitry:
## Issue
Unordered sets can cause unnecessary config changes on every hook and consequently cause unnecessary worker restarts.

## Solution
Sort `peers` set returned by `gather_addresses` that will be used to build `memberslist` configuration


PietroPasotti (approved):
LGTM

PR #87, opened 2024-09-27
S3 tls cert

PietroPasotti:
## Issue
sthe s3 interface may give to the coordinator a "tls-ca-chain" that the worker is expected to use when using the provisioned storage bucket.
At the moment we are not doing anything with it, which means Worker charms can't use storage that is behind tls

Also there is a bad assumption within the coordinator's s3 config, that the "endpoint" we receive via the s3 integration has a scheme prefix. We use that to determine whether insecure=True.
Not only insecure=False will not work given we don't give the worker a tls cert for the storage configuration, but also it turns out the s3 interface doesn't always give us a full url, but could also give us a fqdn (i.e. no scheme).

## Solution
add a field to the cluster schema
coordinator puts there the cert if present in s3 databag, worker picks it up and puts it to filesystem
coordinator uses the cert's presence to determine if `insecure=true` instead of the endpoint scheme.

## Context
too much to tell

## Testing Instructions
added some unittests
we should try to deploy tempo HA with this lib, configure s3 to use a certificate, relate and see if it works.

### DIFF
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -475,7 +475,7 @@ class Coordinator(ops.Object):
             #  tls_server_name: Path to the CA certificate file.  # not a typo: it's the same
             # we send to the worker the chain itself, but the tempo config actually wants a path to a file
             # the worker will be responsible for putting it there
-            "tls_cert_path": worker.S3_TLS_CA_CHAIN_FILE if s3_data.tls_ca_chain else None,
+            "tls_ca_path": worker.S3_TLS_CA_CHAIN_FILE if s3_data.tls_ca_chain else None,
         }
 
         return s3_config
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -727,7 +727,7 @@ class Coordinator(ops.Object):
                 if self.remote_write_endpoints_getter
                 else None
             ),
-            s3_tls_ca_cert=self._s3_config,
+            s3_tls_ca_chain=self.s3_connection_info.tls_ca_chain,
         )
 
     def _render_workers_alert_rules(self):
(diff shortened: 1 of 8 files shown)


sed-i:
Need to remember to release and bump coordinators' cos-lib.

src/cosl/coordinated_workers/coordinator.py
```diff
@@ -434,24 +465,24 @@ def _s3_config(self) -> Dict[str, Any]:
...
-        s3_config["bucket_name"] = s3_data.pop("bucket")
+        s3_data = self.s3_connection_info
+        s3_config = {
+            "endpoint": re.sub(rf"^{urlparse(s3_data.endpoint).scheme}://", "", s3_data.endpoint),
+            "region": s3_data.region,
+            "access_key_id": s3_data.access_key,
+            "secret_access_key": s3_data.secret_key,
+            "bucket_name": s3_data.bucket,
+            "insecure": not s3_data.tls_ca_chain,
+            # FIXME: s3 gives us a cert chain. tempo's s3 config takes:
```
mmkay, line 476: [now that the worker does it](https://github.com/canonical/cos-lib/pull/87/files#diff-2730d74704f2579d8d7f8f29dd820ba64be312efd1aa5053ec464bc02ee12675R564) is this `FIXME`  still needed?

src/cosl/coordinated_workers/worker.py
```diff
@@ -548,71 +547,68 @@ def _update_worker_config(self) -> bool:
...
+                if current_contents == new_contents:
+                    logger.debug(f"{file} unchanged")
+                    continue
+
+            logger.debug(f"{file} updated")
+            any_changes = True
+            self._container.push(file, new_contents, make_dirs=True)
+
+        # Save the cacert in the charm container for charm traces
+        if tls_data.ca_cert:
```
mmkay, line 587: This could also be checked for changes before writing.

mmkay (approved):

src/cosl/coordinated_workers/worker.py
PietroPasotti: I don't think it's worth the extra complexity, it's not a lot of IO

PR #91, opened 2024-10-11
Add guard against coordinator publishing worker config

michaeldmitry:
## Issue
A coordinator should not publish the config to a worker if a worker has not yet published its own data (like roles and addresses).

## Solution
Add a guard to check if the worker has already published its data right before we try and send the config to that worker.

### DIFF
--- a/src/cosl/coordinated_workers/interface.py
+++ b/src/cosl/coordinated_workers/interface.py
@@ -289,7 +289,7 @@ class ClusterProvider(Object):
         for relation in self._relations:
             if relation:
                 if not self._has_worker_published(relation):
-                    log.debug("Worker %s hasn't yet published its data", relation.id)
+                    log.debug("Worker %s hasn't yet published its data", relation.app.name)
                     continue
                 local_app_databag = ClusterProviderAppData(
                     worker_config=worker_config,
@@ -407,24 +407,26 @@ class ClusterProvider(Object):
         return None
 
     def _has_worker_published(self, relation: ops.Relation) -> bool:
-        """Verify that the worker has published its data.
+        """Verify that each worker unit and the worker leader have published their data to the cluster relation.
 
         - unit address is published
         - roles are published
         """
-        if not relation.app:
+        if not relation.app or not relation.units or not relation.data:
             return False
-        if len(relation.units) == 0:
-            return False
+
+        # check if unit data is published
         for worker_unit in relation.units:
             try:
                 ClusterRequirerUnitData.load(relation.data[worker_unit])
-                if self._charm.unit.is_leader():
-                    ClusterRequirerAppData.load(relation.data[relation.app])
-            except DataValidationError as e:
-                log.info(
-                    f"invalid databag contents while checking if worker has published data: {e}"
-                )
+            except DataValidationError:
+                return False
+
+        # check if app data is published
+        if self._charm.unit.is_leader():
+            try:
+                ClusterRequirerAppData.load(relation.data[relation.app])
+            except DataValidationError:
                 return False
 
         return True
--- a/src/cosl/coordinated_workers/interface.py
+++ b/src/cosl/coordinated_workers/interface.py
@@ -288,6 +288,9 @@ class ClusterProvider(Object):
         """Publish the config to all related worker clusters."""
         for relation in self._relations:
             if relation:
+                if not self._has_worker_published(relation):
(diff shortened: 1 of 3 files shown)


src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,29 @@ def get_address_from_role(self, role: str) -> Optional[str]:
...
+            return False
+        if len(relation.units) == 0:
+            return False
+        for worker_unit in relation.units:
+            try:
+                ClusterRequirerUnitData.load(relation.data[worker_unit])
+                if self._charm.unit.is_leader():
+                    ClusterRequirerAppData.load(relation.data[relation.app])
+            except DataValidationError as e:
+                log.info(
```
PietroPasotti, line 425: I'm pretty sure we're already logging this somewhere else at debug level, can you double check that this won't happen twice per hook?

src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,29 @@ def get_address_from_role(self, role: str) -> Optional[str]:
...
+        - roles are published
+        """
+        if not relation.app:
+            return False
+        if len(relation.units) == 0:
+            return False
+        for worker_unit in relation.units:
+            try:
+                ClusterRequirerUnitData.load(relation.data[worker_unit])
+                if self._charm.unit.is_leader():
```
PietroPasotti, line 422: I'd lift this out to the toplevel, we shouldn't be checking for leadership inside a for loop ;)

src/cosl/coordinated_workers/interface.py
PietroPasotti: besides, the point of this method is to check whether the remote end has published valid data. a true/false outcome is enough, I don't think we should be doing any more than that.

src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,29 @@ def get_address_from_role(self, role: str) -> Optional[str]:
             return address_set.pop()
         return None
 
+    def _has_worker_published(self, relation: ops.Relation) -> bool:
+        """Verify that the worker has published its data.
```
PietroPasotti, line 410:
```suggestion
        """Verify that each worker unit and the worker leader have published their data to the cluster relation.
```

src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,29 @@ def get_address_from_role(self, role: str) -> Optional[str]:
...
+
+        - unit address is published
+        - roles are published
+        """
+        if not relation.app:
```
PietroPasotti, line 415:
```suggestion
        if not relation.app or not relation.units or not relation.data:  
```

src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,29 @@ def get_address_from_role(self, role: str) -> Optional[str]:
...
+            return False
+        if len(relation.units) == 0:
+            return False
+        for worker_unit in relation.units:
+            try:
+                ClusterRequirerUnitData.load(relation.data[worker_unit])
+                if self._charm.unit.is_leader():
+                    ClusterRequirerAppData.load(relation.data[relation.app])
```
PietroPasotti, line 423: and actually we shouldn't be doing this either inside the for loop. it's enough to check app data once.

PietroPasotti (changes requested):
some logic tweaks in the new methods, the rest looks good!

src/cosl/coordinated_workers/interface.py
```diff
@@ -403,6 +406,31 @@ def get_address_from_role(self, role: str) -> Optional[str]:
             return address_set.pop()
         return None
 
+    def _has_worker_published(self, relation: ops.Relation) -> bool:
```
PietroPasotti, line 409: [nit] I'm wondering if `_remote_data_ready` would be a better name for this

src/cosl/coordinated_workers/interface.py
```diff
@@ -288,6 +288,9 @@ def publish_data(
         """Publish the config to all related worker clusters."""
         for relation in self._relations:
             if relation:
```
PietroPasotti, line 290: I'd do:
```suggestion
            if relation and self._remote_data_ready(relation):
```

PietroPasotti (approved):
LGTM, couple more minor nits/thoughts

PR #92, opened 2024-10-24
Use pebble notices instead of retrying until the workload is up.

michaeldmitry:
## Issue
We have extra complexity of adding retries using `tenacity` to wait until the workload is `ready`. We also block the hook execution queue until its up or we've exhausted all retries.

## Solution
A better solution would be to replace the retry logic with pebble notices to unblock the hook execution and be notified with a custom pebble notice event when our workload is up. Then, we could reflect that in our charm's status.

We'll define a pebble service `workload-ready` that is disabled by default. When we try to start/restart the pebble workload service, we'll also start the `workload-ready` service. What that will do is keep trying to ping our pebble check `ready` which will report if the workload is ready or not. Once that is up, `workload-ready` will fire a pebble notice which will be observed by our worker to stop that service (to prevent a continuous unnecessary loop of "up" notices). 

## Future enhancements
Currently, the way we implement notices is to notify that the workload is up after it has started, but doesn't fire notices when the workload goes down. 

## Testing Considerations
1. `git clone https://github.com/canonical/tempo-worker-k8s-operator.git`
2. `git clone https://github.com/canonical/tempo-coordinator-k8s-operator.git`
3. `cd tempo-worker-k8s-operator`
4. in `requirements.txt`, replace `cosl>=xx` with `cosl@git+https://github.com/canonical/cos-lib@OPENG-2871`
5. add `git` as part of the `build-packages` in `charmcraft.yaml`
6. `charmcraft pack`
7. cd `tempo-coordinator-k8s-operator`
8. run `TEMPO_WORKER_CHARM=/path/to/tempo-worker-k8s_ubuntu-22.04-amd64.charm tox -vve integration`

### DIFF
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -268,7 +268,7 @@ class Worker(ops.Object):
                         # Fire the pebble notice when pebble "ready" check reports "up".
                         # The assumption here is that pebble "ready" check would keep reporting "up"
                         # until it witnesses 3 consecutive failures, where 3 is set by the "threshold" field in the check layer.
-                        # So, we wait for 30 seconds before checking if its up so that we gurantee that the check is
+                        # So, we wait for 30 seconds before checking if its up so that we guarantee that the check is
                         # reporting the correct status.
                         "command": f"""bash -c 'while true; do sleep 30; /charm/bin/pebble checks ready | grep up &&
                                    /charm/bin/pebble notify {self.ready_notice_key} ||
--- a/src/cosl/coordinated_workers/worker.py
+++ b/src/cosl/coordinated_workers/worker.py
@@ -19,7 +19,7 @@ import tenacity
 import yaml
 from ops import MaintenanceStatus, StatusBase
 from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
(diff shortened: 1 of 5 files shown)


src/cosl/coordinated_workers/worker.py
```diff
@@ -88,19 +89,14 @@ class ServiceEndpointStatus(Enum):
 class Worker(ops.Object):
     """Charming worker."""
 
+    ready_notice_key = "canonical.com/workload-ready"
```
mmkay, line 92: Nit: this could be a constant.

src/cosl/coordinated_workers/worker.py
```diff
@@ -203,6 +203,24 @@ def __init__(
         )
 
     # Event handlers
+    def _on_pebble_custom_notice(self, event: ops.PebbleNoticeEvent):
+        if event.notice.key == self.ready_notice_key:
+            # collect-unit-status should now report ready.
+            logger.debug("pebble API reports ready")
```
mmkay, line 209: Nit: still needed? Doesn't provide lots of context IMO.

mmkay (approved):
A few small nits but otherwise it looks right :)

PietroPasotti:
Decided to drop this and instead wait for juju 3.6 stable so we can use the check-failed and check-recovered events to achieve the same.

PR #93, opened 2024-11-06
added guard for permission errors on relation-get

PietroPasotti:
workers may fail because they try to access cluster relation data on `install`, but something doesn't let them. 
This PR adds a corresponding guard and a log for what should be a transient error.
This behaviour is hard to verify in unittests as the backends don't raise this error (as this is probably an issue in juju)

flyby: minor refactoring of the status-check logic


src/cosl/coordinated_workers/interface.py
```diff
@@ -523,7 +523,16 @@ def publish_app_roles(self, roles: Iterable[str]):
...
+            try:
+                databag_model.dump(relation.data[self.model.app])
+            except ModelError as e:
+                if "ERROR permission denied (unauthorized access)" in e.args:
+                    # if we are handling an event prior to 'start', we could be denied write access
+                    # hopefully we can ignore it and trust that it will be retried at the next occasion
+                    log.debug(
+                        "databag dump gave a permission denied error. "
+                        "This could be a transient issue."
+                    )
```
sed-i, line 535: If we think swallowing this error is a good idea, 
- should it be done inside `.dump()`?
- substring check is brittle - I think we can safely log a `debug` for any ModelError.

But I think that this approach may bite us: model errors are not meant to be dropped - worst case we'd defer the event, but without context at hand I think we should opt for one of the following:
- Caller must guard.
- It's ok to go to error state.

Unless I'm missing something?

src/cosl/coordinated_workers/worker.py
```diff
@@ -255,19 +259,26 @@ def status(self) -> ServiceEndpointStatus:
         # we really don't want this code to raise errors, so we blanket catch all.
         try:
             services = self._container.get_services(*layer.services.keys())
-            running_status = {name: svc.is_running() for name, svc in services.items()}
-            if not all(running_status.values()):
-                if any(running_status.values()):
-                    starting_services = tuple(
+            if services:
+                # avoid vacuous quantification
```
sed-i, line 263: Should we bring this comment a few [reading levels](https://hemingwayapp.com/) down? :sweat_smile: 
That being said, the new comment seems to be a pull request comment rather than code comment.

src/cosl/coordinated_workers/worker.py
```diff
@@ -255,19 +259,26 @@ def status(self) -> ServiceEndpointStatus:
...
+                running_status = {name: svc.is_running() for name, svc in services.items()}
```
sed-i, line 264: Given the extra logic below, should we replace this with a plain for that populates two separate dicts, `active`, `stopped`? Perhaps an ops feature request?

src/cosl/coordinated_workers/interface.py
```diff
@@ -523,7 +523,16 @@ def publish_app_roles(self, roles: Iterable[str]):
...
             databag_model = ClusterRequirerAppData(role=",".join(roles))
-            databag_model.dump(relation.data[self.model.app])
+            try:
+                databag_model.dump(relation.data[self.model.app])
+            except ModelError as e:
+                if "ERROR permission denied (unauthorized access)" in e.args:
+                    # if we are handling an event prior to 'start', we could be denied write access
+                    # hopefully we can ignore it and trust that it will be retried at the next occasion
+                    log.debug(
+                        "databag dump gave a permission denied error. "
```
mmkay, line 533:
```suggestion
                        "Saving contents of databag failed with a permission denied error."
```

src/cosl/coordinated_workers/worker.py
PietroPasotti: it was in fact meant to be a code comment, explaining why we need the `if(services)` check to verify it's not empty before doing an `all` on it, but I'll rephrase

src/cosl/coordinated_workers/worker.py
PietroPasotti: I wouldn't mind it being in ops, but I'm not sure it's general enough to make the bar. will see if the split makes the logic more clear

PietroPasotti: turned out flipping the polarity of the check made the logic simpler, we don't need to track the 'started' services at all.

src/cosl/coordinated_workers/interface.py
PietroPasotti: I like the idea of letting the caller guard, so the Worker will have to  do that in `_update_cluster_relation`.
afraid we can't make the substring check any less brittle as that is all we get from ops.

as for the error state, that should be only if we get a different error, one we don't trust is transient.
Updated.

mmkay (approved):

src/cosl/coordinated_workers/worker.py
```diff
@@ -501,7 +508,19 @@ def _update_cluster_relation(self) -> None:
         self.cluster.publish_unit_address(socket.getfqdn())
         if self._charm.unit.is_leader() and self.roles:
             logger.info(f"publishing roles: {self.roles}")
-            self.cluster.publish_app_roles(self.roles)
+            try:
+                self.cluster.publish_app_roles(self.roles)
+            except ModelError as e:
+                # if we are handling an event prior to 'start', we could be denied write access
+                # hopefully we can ignore it and trust that it will be retried at the next occasion
```
sed-i, line 515: If you really want to do this here, then I'd rather compare envvar against HOOK_INSTALL or something.

In any case, I still think it's better to let the caller guard. When we make such antipattern decisions deep down, production ends up breaking in weird ways, calling for re-relate, refresh "fixes".

PietroPasotti: Thing is there is no hook we can compare against, since the behaviour is "error if we haven't seen install **yet**" and we cannot look into the future. The only solution would be store a history of the events we've seen so far, but that feels un-opsy. So this is the only stateless solution I see.

As for the 'caller should guard' principle, I'm on board, but Worker _is_ the caller.

PR #97, opened 2024-11-11
chore: use ops[testing] rather than ops-scenario<7

tonyandrewmeyer:
## Issue

cos-lib currently pins ops-scenario to <7, which means that any charms that are using cos-lib are also stuck with that version of scenario.

## Solution

Updates to the latest published version of ops-scenario, and also to installing with `ops[testing]` and using from the `ops.testing` namespace.

## Context

Latest scenario/ops.

## Testing Instructions

`tox`

## Upgrade Notes

N/A, only tests are changed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "cosl"
-version = "0.0.42"
+version = "0.0.43"
 authors = [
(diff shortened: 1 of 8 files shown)


PietroPasotti (approved):
thanks for this! LGTM. the integration test PRs fail because my actions aren't ready to do cross-repo checkouts

PR #102, opened 2024-11-18
fix: Adding test_invalid_databag_content for ClusterProvider

mathmarchand:
## Issue
Missing test coverage in src/coodinated_workers/interface.py around 335-337, 342-344.

## Solution
Create an additional test and passing invalid databag for both the ClusterRequiredApp and ClusterRequiredUnit, then check logs output for "invalid databag" errors.

## Context

The ClusterProvider object is a relation endpoint wrapper for the 'cluster' relation interface.
That piece of code is responsible for walking through all units related over 'cluster' and collect a piece of data from their databags.

## Testing Instructions
run `tox -e unit`

## Upgrade Notes
No impacts on the code itself, just adding a additional test coverage.

### DIFF
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -93,7 +93,7 @@ class S3ConnectionInfo(pydantic.BaseModel):
     access_key: str = pydantic.Field(alias="access-key")  # type: ignore
     secret_key: str = pydantic.Field(alias="secret-key")  # type: ignore
 
-    region: Optional[str] = pydantic.Field(None)
+    region: Optional[str] = pydantic.Field(None)  # type: ignore
     tls_ca_chain: Optional[List[str]] = pydantic.Field(None, alias="tls-ca-chain")  # type: ignore
 
     @property


tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,103 @@ def test_tracing_receivers_urls(
...
+        containers={
+            testing.Container("nginx", can_connect=True),
+            testing.Container("nginx-prometheus-exporter", can_connect=True),
+        },
+        relations=list(requires_relations.values()) + list(provides_relations.values()),
+    )
+
+
+@pytest.fixture()
+def cluster_charm(request):
```
PietroPasotti, line 344: why can't you reuse the `coordinator_charm` fixture instead of adding this very similar one?
also, I don't think you need the `request` param

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,103 @@ def test_tracing_receivers_urls(
             "otlp_http": "5.6.7.8:4318",
             "otlp_grpc": "5.6.7.8:4317",
         }
+
+
+@pytest.fixture
+def cluster_state():
```
PietroPasotti, line 290: since you're not reusing this fixture, and it's not very generic, I'd recommend putting the code directly in the test. 
That or, give it a more descriptive name such as "state_with_cluster_relation_invalid_databag"

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,103 @@ def test_tracing_receivers_urls(
...
+                charm=self,
+                # Roles were take from loki-coordinator-k8s-operator
+                roles=frozenset({"all", "read", "write", "backend"}),
+                endpoint="my-logging",
+            )
+
+    return MyCluster
+
+
+def test_invalid_databag_content(
```
PietroPasotti, line 379: as it is this test is rather brittle: in general it's not a great practice to test logging output as that is subject to changing.
Is there something else we can assert to verify the right behaviour?
IIRC the charm will set a Waiting/Blocked status if the remote databag is invalid. That is more robust, and does not require you to use the context manager and poke into the charm's internals (which are also brittle)

PietroPasotti (changes requested):
Hi there! thanks for this. A few comments on how to make the test more robust.

tests/test_coordinated_workers/test_coordinator.py
mathmarchand: I can, I have switch to using it.

tests/test_coordinated_workers/test_coordinator.py
mathmarchand: I have moved it to the test function.

tests/test_coordinated_workers/test_coordinator.py
mathmarchand: I have switch to check the unit status, but using MyCoordinator CluserProvider was a bit complexe and calling the function directly still seems the best. I am opened to look further, but would require a bit of your help to understand how to acheive it.

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,46 @@ def test_tracing_receivers_urls(
             "otlp_http": "5.6.7.8:4318",
             "otlp_grpc": "5.6.7.8:4317",
         }
+
+
+def test_invalid_databag_content(coordinator_charm: ops.CharmBase):
```
PietroPasotti, line 289: Much better! I know we're not super consistent in this ourselves, but in general it's great practice to add comments to your test to clarify what it is that you're testing and separate the concerns of the test itself.
Could you add comments:

```
# GIVEN a coordinator charm with a cluster relation and invalid remote databag contents
... setup code
# WHEN the coordinator processes any event
... ctx
# THEN the coordinator sets blocked
... assert
```

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,46 @@ def test_tracing_receivers_urls(
...
+        containers={
+            testing.Container("nginx", can_connect=True),
+            testing.Container("nginx-prometheus-exporter", can_connect=True),
+        },
+        relations=list(requires_relations.values()) + list(provides_relations.values()),
+    )
+
+    ctx = testing.Context(coordinator_charm, meta=coordinator_charm.META)
+    with ctx(ctx.on.start(), invalid_databag_state) as manager:
+        cluster = ClusterProvider(manager.charm, roles=frozenset(["all"]), endpoint="my-logging")
```
PietroPasotti, line 327:
```suggestion
        cluster = manager.charm.coordinator.cluster
```

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,46 @@ def test_tracing_receivers_urls(
...
+    invalid_databag_state = testing.State(
+        containers={
+            testing.Container("nginx", can_connect=True),
+            testing.Container("nginx-prometheus-exporter", can_connect=True),
+        },
+        relations=list(requires_relations.values()) + list(provides_relations.values()),
+    )
+
+    ctx = testing.Context(coordinator_charm, meta=coordinator_charm.META)
+    with ctx(ctx.on.start(), invalid_databag_state) as manager:
```
PietroPasotti, line 326: to clarify that we're not testing the 'start' event specifically, but any, you could parametrize and add a couple more. For example:

```python
@pytest.mark.parametrize("event",
(
scenario.CharmEvents.update_status(),
scenario.CharmEvents.start(),
scenario.CharmEvents.install(),
) 
)
```

PietroPasotti (changes requested):
looking better! a few more fixes necessary, but you're on the right track

tests/test_coordinated_workers/test_coordinator.py
mathmarchand: Not sure why, but if I call the cluster that way, the function gather_addresses_by_role does not get called anymore.

mathmarchand: Nevermind found out why.

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -281,3 +284,46 @@ def test_tracing_receivers_urls(
...
+        cluster = ClusterProvider(manager.charm, roles=frozenset(["all"]), endpoint="my-logging")
+        cluster.gather_addresses_by_role()
+        assert cluster.model.unit.status == ops.UnknownStatus()
```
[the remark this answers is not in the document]
mathmarchand: Switch to using your suggestion, and checked for blocked status on the unit.

PietroPasotti (approved):
fantastic! thanks a lot

PietroPasotti:
@mathmarchand I'd like to merge but could you sign your commits?

mathmarchand:
@PietroPasotti the commit have been signed. Thanks for your help!

PR #105, opened 2024-11-21
[wip] catch charm tracing config error

PietroPasotti:
## Issue
There's a race condition with TLS and ingress.

When a worker is related to ingress, and you add a TLS provider:

- If it first sees a certificates-changed and receives the cert, and THEN gets an https endpoint, things are fine.
- However, if it first sees an ingress-changed and receives an https endpoint then it will go in error state because it doesn't have a cert yet.

Juju sees the error state and refuses to let the charm state progress, so the charm is stuck in error state.
(Note that this last bit is theoretical, in practice the charm somehow snapped out of error state after a couple of retries, but I have no idea why.)

Either way, should we allow the charm to go into error state or should we catch it and set blocked, informing the user that the issue MIGHT resolve itself assuming they integrated TLS correctly to both tempo and ingress? And that, if it doesn't resolve itself within X events, then they are probably missing an integration between certs, tempo and ingress?

Note that we don't **know** if the status should be blocked or waiting, because it could be that we're just waiting for an event, but it could also be that we're missing an integration (between two OTHER charms) and the event will never come.

@ reviewers: (don't focus on the implementation for now, focus on the problem)


sed-i:
Should we somehow plug an `_is_tls_ready()` that will give True when all conditions are met, and until then the tracing config itself will have plain http? I imagine failing to push to the backend for a bit is ok?

Related: https://github.com/canonical/loki-k8s-operator/pull/416

src/cosl/coordinated_workers/worker.py
```diff
@@ -340,6 +349,18 @@ def _on_collect_status(self, e: ops.CollectStatusEvent):
             statuses.append(BlockedStatus("Missing relation to a coordinator charm"))
         elif not self.cluster.relation:
             statuses.append(WaitingStatus("Cluster relation not ready"))
+
+        try:
+            self._charm_tracing_endpoint, self._charm_tracing_cert = self.charm_tracing_config()
+        except CharmTracingTLSConfigurationError:
```
sed-i, line 355: Can we avoid the raise-and-swallow by changing the approach underneath?

PietroPasotti: as I said in the description, let's not focus on the implementation just yet

tonyandrewmeyer:
> Either way, should we allow the charm to go into error state or should we catch it and set blocked, informing the user that the issue MIGHT resolve itself assuming they integrated TLS correctly to both tempo and ingress? And that, if it doesn't resolve itself within X events, then they are probably missing an integration between certs, tempo and ingress?

I'm a fan of using error state when it makes sense, but I don't think this is a good case. I think catch-and-(blocked?) is the way to go, because it does describe the problem: the charm is blocked by / waiting for something, it's not really broken in the way that an error would indicate.

> Note that we don't **know** if the status should be blocked or waiting, because it could be that we're just waiting for an event, but it could also be that we're missing an integration (between two OTHER charms) and the event will never come.

Yeah, this part is tricky. I think on balance this means that blocked is the better choice, because it *might* be that human intervention is required, so better to let someone know. If it should have actually been waiting and resolves itself in the meantime, then that's not as bad as if it was set to waiting but was actually needing someone to integrate and no-one does because they don't know.

PR #107, opened 2024-11-26
Added unit tests for ClusterRequirer interface class

lorenzo-medici:
## Issue
Missing test coverage for `src/coordinated_workers/interface.py` in class `ClusterRequirer`

## Solution
Create additional tests with invalid databags, specifically when the cluster relation changes

## Context
The ClusterProvider object is a relation endpoint wrapper for the 'cluster' relation interface.
That piece of code is responsible for walking through all units related over 'cluster' and collect a piece of data from their databags.

## Testing Instructions
Run `tox -e unit`

## Upgrade Notes
No code changes, only effect is increased test coverage.


tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -344,3 +347,72 @@ def test_invalid_databag_content(coordinator_charm: ops.CharmBase, event):
...
+            "containers": {"foo": {"type": "oci-image"}},
+        },
+        config={"options": {"role-all": {"type": "boolean", "default": True}}},
+    )
+
+    # AND the relation changes
+    # IF the relation data is invalid (forced by the patched Exception)
+    relation = testing.Relation(
+        "cluster",
+        remote_app_data={"worker_config": json.dumps("some: yaml")},
```
PietroPasotti, line 373: I believe that this line is unnecessary, since the cluster provider app data that you are checking is the local side of the relation (from the pov of the coordinator).
Also, worker_config is a field in the coordinator's databag, so it should be on 'this side'.

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -344,3 +347,72 @@ def test_invalid_databag_content(coordinator_charm: ops.CharmBase, event):
...
+    )
+
+    # THEN the charm emits a ClusterRemovedEvent
+    with ctx(ctx.on.relation_changed(relation),
```
PietroPasotti, line 377: since you're not using the manager at all, this should be replaced with the regular `ctx.run()` flow

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -344,3 +347,72 @@ def test_invalid_databag_content(coordinator_charm: ops.CharmBase, event):
...
+
+    # AND the relation changes
+    # IF the relation data is invalid (forced by the patched Exception)
+    relation = testing.Relation(
+        "cluster",
+        local_unit_data={"new": "config"},
+    )
+
+    # THEN the charm only emits a RelationChangedEvent
+    with ctx(ctx.on.relation_changed(relation),
```
PietroPasotti, line 412: same as above

tests/test_coordinated_workers/test_coordinator.py
```diff
@@ -344,3 +347,72 @@ def test_invalid_databag_content(coordinator_charm: ops.CharmBase, event):
...
+            "containers": {"foo": {"type": "oci-image"}},
+        },
+        config={"options": {"role-all": {"type": "boolean", "default": True}}},
+    )
+
+    # AND the relation changes
+    # IF the relation data is invalid (forced by the patched Exception)
+    relation = testing.Relation(
+        "cluster",
+        local_unit_data={"new": "config"},
```
PietroPasotti, line 408: also probably not necessary. if it is necessary for the test to pass, then you're patching the wrong thing above

PietroPasotti (commented):
In addition to the issues I commented about, if I understand these tests correctly, you might have just found a bug.

I'm not sure why there should be a behavioural difference between 'invalid local app data' and 'invalid local unit data' when it comes to emitting 'cluster removed' events.

Could you mash those two tests into one, parametrize it with an `app` boolean field, depending on which value you choose what to patch and adapt your assertion?
And make a note that the assertion special-casing should not be necessary.

@lucabello do you agree that this is an issue?

Note for @lorenzo-medici: it will be out of scope for you and this PR to fix the bug, I'm just wondering if we should be opening a new one to track or not.

lorenzo-medici:
I merged the two unit tests into one with the `app` parameter making sure to leave a note regarding the difference in behavior.
I also applied the review suggestions.

tests/test_coordinated_workers/test_coordinator.py
lorenzo-medici: Removed in [0005a0b](https://github.com/canonical/cos-lib/pull/107/commits/0005a0b82548eed2448961c87bd4dd4f29bc0486).

tests/test_coordinated_workers/test_coordinator.py
lorenzo-medici: Removed in [0005a0b](https://github.com/canonical/cos-lib/pull/107/commits/0005a0b82548eed2448961c87bd4dd4f29bc0486)

tests/test_coordinated_workers/test_coordinator.py
lorenzo-medici: Changed to `ctx.run()` call in [0005a0b](https://github.com/canonical/cos-lib/pull/107/commits/0005a0b82548eed2448961c87bd4dd4f29bc0486)

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,120 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+        meta={
+            "name": "foo",
+            "requires": {"cluster": {"interface": "cluster"}},
+            "containers": {"foo": {"type": "oci-image"}},
+        },
+        config={"options": {"role-all": {"type": "boolean", "default": True}}},
+    )
+
+    # AND ClusterRequirer is passed an invalid url
+    mock_socket = mock_socket.return_value
```
PietroPasotti, line 797:
```suggestion
    mock_socket.return_value = "213!@#"
```

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,120 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+    mock_socket = "?invalid-url[]"
```
PietroPasotti, line 798: not sure what this line is doing

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,120 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+
+    # IF the charm executes any event
+    # THEN the charm raises an error with the appropriate cause
+    with pytest.raises(UncaughtCharmError) as exc:
+        ctx.run(ctx.on.update_status(), testing.State(containers={testing.Container("foo")}))
+    assert isinstance(exc.value.__cause__, ValueError)
```
PietroPasotti, line 804: TIL.

PietroPasotti (changes requested):
one final nit

tests/test_coordinated_workers/test_worker.py
lorenzo-medici: I found the two lines manipulating the return value of the socket function in a guide, and copied them over. I also wondered why the two lines were necessary, but the test was not passing if I shortcut them as `mock_socket.return_value = "?invalid-url[]"`. I figured that something was going in the mock class and that you first needed to get a reference to the object and then change it - I had the same issue in some other python library in the past.

Since it didn't make much sense and all code examples showed a simple assignment I dug a bit deeper and noticed that the url was not validated but simply parsed, meaning that I was passing a "legitimate" url as far as the parser was concerned.
The manipulation in these two lines simply broke the mock class.
So simply picking an url that triggers an exception in the `urlparse` method solved the problem, in this case one with mismatched square brackets: `http://www.invalid-]url.com`.

(I couldn't simply accept your suggestion because the url you proposed didn't break the parser either)

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,119 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+    relation = testing.Relation(
+        "cluster",
+        remote_app_data=remote_databag,
+    )
+    with ctx(
+        ctx.on.relation_changed(relation),
+        testing.State(containers={container}, relations={relation}),
+    ) as mgr:
+        charm = mgr.charm
+        mgr.run()
```
PietroPasotti, line 846: I'm not sure in this test and the next you need to mgr.run() before your assertion. Could you check whether that is necessary?
The difference is subtle: mgr.run() emits the event, so the question is: does the return value of `get_charm_tracing_receivers` depend on the charm instance state? I hope and suspect that it shouldn't, so we can write the assertion without emitting the event first.

This also applies to the test below

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,119 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+    container = testing.Container(
+        "foo",
+        execs={testing.Exec(("update-ca-certificates", "--fresh"))},
+        can_connect=True,
+    )
+    relation = testing.Relation(
+        "cluster",
+        remote_app_data=remote_databag,
+    )
+    with ctx(
```
PietroPasotti, line 841: can you please also add Given/When/Then comments to these two tests?

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,119 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
         stat.value for stat in service_statuses
     ]
+
+
+@patch("socket.getfqdn")
+def test_invalid_url(mock_socket):
+    # Test that when socket returns an invalid url as a Fully Qualified Domain Name,
+    #   ClusterRequirer.publish_unit_address raises a ValueError exception
+
+    # WHEN you define a properly configured charm
```
PietroPasotti, line 785: this is the GIVEN step

tests/test_coordinated_workers/test_worker.py
```diff
@@ -774,3 +775,119 @@ def test_worker_stop_all_services_if_not_ready(tmp_path):
...
+            "requires": {"cluster": {"interface": "cluster"}},
+            "containers": {"foo": {"type": "oci-image"}},
+        },
+        config={"options": {"role-all": {"type": "boolean", "default": True}}},
+    )
+
+    # AND ClusterRequirer is passed an invalid url
+    mock_socket.return_value = "http://www.invalid-]url.com"
+
+    # IF the charm executes any event
```
PietroPasotti, line 799: and this is a WHEN

PietroPasotti (changes requested):
much better! last pass :)

tests/test_coordinated_workers/test_worker.py
lorenzo-medici: The test still pass with the assertions before mgr.run(). Should I execute mgr.run() anyways or remove it as it is unnecessary?

PietroPasotti: remove it

lorenzo-medici:
Is this failure my fault? All tests pass locally, and the failure point in the GH workflow is 
```
/home/runner/work/cos-lib/cos-lib/src/charm.py
  /home/runner/work/cos-lib/cos-lib/src/charm.py:27:48 - error: "DatabagModel" is not exported from module "cosl.coordinated_workers.interface"
    Import from "cosl.interfaces.utils" instead (reportPrivateImportUsage)
```
which I don't know how to fix as `DatabagModel` is always imported from `cosl.interfaces.utils`.

PietroPasotti:
> Is this failure my fault? All tests pass locally, and the failure point in the GH workflow is
> 
> ```
> /home/runner/work/cos-lib/cos-lib/src/charm.py
>   /home/runner/work/cos-lib/cos-lib/src/charm.py:27:48 - error: "DatabagModel" is not exported from module "cosl.coordinated_workers.interface"
>     Import from "cosl.interfaces.utils" instead (reportPrivateImportUsage)
> ```
> 
> which I don't know how to fix as `DatabagModel` is always imported from `cosl.interfaces.utils`.

No that's a recent change in cos-lib, it could be that the tempo coordinator's integration tests are broken because of it.

PietroPasotti (approved):

PR #110, opened 2024-12-10
directly vendor schema files

PietroPasotti:
## Issue
Pypi packages apparently cannot depend on git+ dependencies.
https://github.com/canonical/cos-lib/actions/runs/12252877996/job/34180522427

[charm-relation-interfaces is not on pypi (yet)
](https://github.com/canonical/charm-relation-interfaces/issues/213)

## Solution
Instead of installing 'interfaces' from git, download them in this repository tree for the time being.
This is a short term workaround, not a solution.
If charm-relation-interfaces is not to be on pypi, a better approach would be to pull the schemas automatically from CI every time there's a merge on CRI main.


lucabello (approved):

NucciTheBoss:
@PietroPasotti what was the reason for adding `cryptography` as a dependency in _pyproject.toml_? As of the latest `cosl` release, every charm that pulls in `cosl` will need to include a Rust toolchain during the packing process to compile `cryptography`, or add an entry for `cryptography` in `charm-binary-python-packages`.

PR #117, opened 2025-01-15
Centralized rules groups for libraries to import

MichaelThamm:
This PR attempts to centralize the generic alert rule groups in cos-lib to remove code duplication. It is an extension of this merged PR:
- https://github.com/canonical/cos-lib/pull/115

Tandem PRs:
- https://github.com/canonical/prometheus-k8s-operator/pull/660
- https://github.com/canonical/grafana-agent-operator/pull/232
- https://github.com/canonical/avalanche-k8s-operator/pull/50

Note:
- This will not break any CI since the generic rules existed in the Prometheus and Grafana agent libraries. Those will still exist there until the migration to these rules in `cosl` is complete.
- This PR is not dependent on the status of the tandem PRs, rather they are dependent on this one!

### DIFF
--- a/src/cosl/rules.py
+++ b/src/cosl/rules.py
@@ -96,6 +96,8 @@ logger = logging.getLogger(__name__)
 
 
 _generic_alert_rules: Final = SimpleNamespace(
+    # We use "5m" to avoid false positives on expected temporary "down", e.g. during intentional (re)start.
+    # Juju topology will be later injected by providers of alert rules.
     host_down={
         "alert": "HostDown",
         "expr": "up < 1",
@@ -110,6 +112,7 @@ _generic_alert_rules: Final = SimpleNamespace(
     },
     host_metrics_missing={
         "alert": "HostMetricsMissing",
+        # We use "for: 5m" instead of "absent_over_time" because ... (I forget)
         "expr": "absent(up)",
         "for": "5m",
         "labels": {"severity": "critical"},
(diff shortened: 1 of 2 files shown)


src/cosl/rules.py
```diff
@@ -94,6 +95,63 @@
 logger = logging.getLogger(__name__)
 
 
+_generic_alert_rules: Final = SimpleNamespace(
+    host_down={
```
sed-i, line 99:
```suggestion
    # We use "5m" to avoid false positive on expected temporary "down", e.g. during intentional (re)start.
    host_down={
```

src/cosl/rules.py
```diff
@@ -94,6 +95,63 @@
...
+        "alert": "HostDown",
+        "expr": "up < 1",
```
sed-i, line 101:
```suggestion
        # Juju topology will be later injected by providers of alert rules.
        "expr": "up < 1",
```

src/cosl/rules.py
```diff
@@ -94,6 +95,63 @@
...
+        "annotations": {
+            "summary": "Host '{{ $labels.instance }}' is down.",
+            "description": """Host '{{ $labels.instance }}' is down, failed to scrape.
+                            VALUE = {{ $value }}
+                            LABELS = {{ $labels }}""",
+        },
+    },
+    host_metrics_missing={
+        "alert": "HostMetricsMissing",
+        "expr": "absent(up)",
```
sed-i, line 113:
```suggestion
        # We use "for: 5m" instead of "absent_over_time" because ... (I forget)
        "expr": "absent(up)",
```

src/cosl/rules.py
```diff
@@ -94,6 +95,63 @@
...
+        },
+    },
+)
+
+"""
+Generic alert rules are in groups to ensure a predictable group name. The group names must be unique in a Prometheus installation.
+Charmed Prometheus-k8s handles this with injecting JujuTopology information into the group name.
+"""
+generic_alert_groups: Final = SimpleNamespace(
+    application_rules={
```
sed-i, line 130:
```suggestion
    # Group names must be unique per alert rule file. The final group names may be adjusted by
    # the providers of alert rules to include some topology information, to addresses deduplication.
    application_rules={
```

sed-i (approved):
Nice! Added some doc suggestions.

PR #128, opened 2025-02-18
Add a utility function to check for ipv6

michaeldmitry:
## Issue
Our HA solutions running nginx insert a default config for listening on ipv6 addresses. In an environment where ipv6 is disabled, nginx fails to start.

## Solution
Add a common utility function that can be used by the nginx config to add/remove ipv6-related options.

## Testing Instructions
Tandem PR: TODO

### DIFF
--- a/src/cosl/coordinated_workers/nginx.py
+++ b/src/cosl/coordinated_workers/nginx.py
@@ -4,6 +4,7 @@
 """Workload manager for Nginx. Used by the coordinator to load-balance and group the workers."""
 
 import logging
+import subprocess
 from pathlib import Path
 from typing import Callable, Optional, TypedDict
 
@@ -205,3 +206,15 @@ class NginxPrometheusExporter:
                 },
             }
         )
+
+
+def is_ipv6_enabled() -> bool:
+    """Check if IPv6 is enabled on the container's network interfaces."""
+    exit_code, output = subprocess.getstatusoutput("ip -6 address show")
+    if exit_code != 0:
+        # if running the command failed for any reason, assume ipv6 is not enabled.
+        logger.warning(f"Disabling ipv6 in nginx config: {output}")
+        return False
+    if output:
+        return True
+    return False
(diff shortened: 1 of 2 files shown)


src/cosl/coordinated_workers/nginx.py
```diff
@@ -205,3 +206,15 @@ def layer(self) -> pebble.Layer:
                 },
             }
         )
+
+
+def is_ipv6_enabled() -> bool:
+    """Check if IPv6 is enabled on the container's network interfaces."""
+    exit_code, output = subprocess.getstatusoutput("ip -6 address show")
```
PietroPasotti, line 213: I'd use `run` instead; it has a bunch of handy args for much of what you're manually doing below:

```
def is_ipv6_enabled():
    try:
        out = subprocess.run(["ip", "-6", "address", "show"], check=True, capture_output=True, text=True)
    except:
        return False
    return bool(out)
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -205,3 +206,15 @@ def layer(self) -> pebble.Layer:
...
+    if exit_code != 0:
+        # if running the command failed for any reason, assume ipv6 is not enabled.
+        logger.warning(f"Disabling ipv6 in nginx config: {output}")
```
PietroPasotti, line 216: I think it's up to the caller to log this. this function doesn't know why it's being called.

PietroPasotti (changes requested):
I wonder if there's anything else but this utility that we should be baking into cosl.
Should we be wiring the nginx config generation to do anything with this information? Or is it up to the coordinator charms to use this utility and adapt their config accordingly?

michaeldmitry:
@PietroPasotti issue is the changes need for disabling/enabling ipv6 config-related stuff is too low on the nginx config. So, I think its better to leave the config manipulation where it already is (i.e each coordinator should handle on their own)

PietroPasotti (approved):
LGTM!

PR #130, opened 2025-02-21
Publish worker workload versions and the worker config corresponding version

michaeldmitry:
## Issue
Tempo 2.7.1 comes with a breaking change in configuration: configs for <2.7.1 will not work with tempo >=2.7.1 and vice versa.
When doing in-track upgrades, we must find a way to prevent the user from refreshing the worker charms to a revision deploying tempo >=2.7.1, while the coordinator still publishes a <2.7.1 config version or vice versa.

## Solution
- Workers publish their own workload versions to `cluster` app data
- The coordinator would be set to BlockedState if its connected to workers with different workload versions and not give them a config.
- In addition to the config, the coordinator will send `worker_config_version` that corresponds to the workload version that this config is generated for, so that the worker can verify that the coordinator revision is modern enough to be aware of this versioned configuration requirement.

## Testing Instructions
Tandem PRs:
- https://github.com/canonical/tempo-coordinator-k8s-operator/pull/115
- https://github.com/canonical/tempo-worker-k8s-operator/pull/58

### DIFF
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -611,6 +611,14 @@ class Coordinator(ops.Object):
             # if is_recommended is None: it means we don't have a recommended deployment criterion.
             statuses.append(ops.ActiveStatus("Degraded."))
 
+        worker_versions = self.cluster.gather_workload_versions()
+        if len(worker_versions) > 1:
+            statuses.append(
+                ops.BlockedStatus(
+                    f"[consistency] Workers are running different versions: {', '.join(worker_versions)}"
+                )
+            )
+
         if not self.s3_requirer.relations:
             statuses.append(ops.BlockedStatus("[s3] Missing S3 integration."))
         elif not self.s3_ready:
--- a/src/cosl/coordinated_workers/coordinator.py
+++ b/src/cosl/coordinated_workers/coordinator.py
@@ -6,6 +6,7 @@
 import glob
 import json
 import logging
+import math
 import os
 import re
 import shutil
@@ -20,7 +21,9 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Protocol,
     Set,
+    Tuple,
     TypedDict,
 )
 from urllib.parse import urlparse
@@ -53,7 +56,6 @@ check_libs_installed(
     "charms.tls_certificates_interface.v3.tls_certificates",
     "charms.catalogue_k8s.v1.catalogue",
 )
-
 from charms.catalogue_k8s.v1.catalogue import CatalogueConsumer, CatalogueItem
 from charms.data_platform_libs.v0.s3 import S3Requirer
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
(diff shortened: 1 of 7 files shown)


src/cosl/coordinated_workers/coordinator.py
```diff
@@ -236,6 +237,7 @@ def __init__(
             workload_tracing_protocols: A list of protocols that the worker intends to send
                 workload traces with.
             catalogue_item: A catalogue application entry to be sent to catalogue.
+            worker_config_version: A function returning the workload version for which the generated worker config is intended.
```
PietroPasotti, line 240: It feels strange to have this in the constructor.
And why is it a callable?
when the coordinator initializes, we might not yet know what worker config version we need to generate: the worker may still need to tell that to us.

So I'd rather expect the coordinator to expose a `get_requested_worker_version()` method for the caller (the charm) to know what worker version we're connected to

src/cosl/coordinated_workers/coordinator.py
```diff
@@ -370,6 +377,14 @@ def __init__(
             )
             return
 
+        if len(self.cluster.gather_workload_versions()) > 1:
+            logger.error(
+                f"Incoherent deployment. {charm.unit.name} is connected to workers that are running "
+                f"different workload versions: {', '.join(self.cluster.gather_workload_versions())}. "
```
PietroPasotti, line 383: instead of calling that twice, lift it out to a variable or define it in a walrus

src/cosl/coordinated_workers/coordinator.py
PietroPasotti: besides, unless there's precisely one workload version that the workers are requesting, we'll block and do nothing, so there's no point in taking this as an argument from the charm. 
We'll send back the same version we take in, unless we can't satisfy it.

PietroPasotti: so what I think we need instead is:

`self._workers_config_getter = lambda: workers_config(self, version=self.get_requested_worker_version())`

so the charm can implement:

`def generate_worker_config(self, coordinator, version):`

PietroPasotti: may need something smart for that to be backwards compatible, like checking the number of args that the config generator takes

PietroPasotti (changes requested):
I don't see yet the logic for the coordinator to provide different configs based on the requested version.
Is this still a POC?

src/cosl/coordinated_workers/coordinator.py
```diff
@@ -183,6 +185,132 @@ def _validate_container_name(
...
+    def version_range(self) -> VersionRange:
+        """The range of versions that this config builder supports."""
+        ...
+
+    def build(self, coordinator: "Coordinator") -> str:
+        """Build a worker config."""
+        ...
+
+
+class ConfigBuilderFactory:
```
PietroPasotti, line 211: First of all I'm not convinced this is a factory, in the sense of https://realpython.com/factory-method-python/
secondly, in a few lines we have "config_generator", "config_builders", and "configbuilderfactory".

Also, each method of this class contains a switch for the `if self._config_builders` condition, which indicates we should handle this case at a higher level.

I think we need to simplify. 
I'd start by getting rid of the ConfigBuilder abstraction and use the primitive type:
`_VersionedWorkerConfigModels = Dict[VersionRange, Callable[["Coordinator"], str]]`

This comment was 150 lines long when I decided to write a branch instead: 

please diff with `pietros-2c/publish-version` to see what I'm suggesting.

src/cosl/interfaces/cluster.py
```diff
@@ -112,6 +114,8 @@ class ClusterProviderAppData(cosl.interfaces.utils.DatabagModel):
     ### worker node configuration
     worker_config: str
     """The whole worker workload configuration, whatever it is. E.g. yaml-encoded things."""
+    worker_config_version: Optional[str] = None
```
PietroPasotti, line 117: I don't like that we're receiving a `workload_version`, but sending back a `worker_config_version`, although they're supposed to be identical.
I don't see a better alternative though, perhaps `workload_version_supported: true`? In the end the worker isn't going to do anything with that value: all we really want to know is if the coordinator was able to satisfy the version we requested.

src/cosl/interfaces/cluster.py
```diff
@@ -341,6 +346,24 @@ def get_address_from_role(self, role: str) -> Optional[str]:
             return address_set.pop()
         return None
 
+    def gather_workload_versions(self) -> Set[str]:
+        """Gather workers' published workload versions."""
+        data: Set[str] = set()
+        for relation in self._relations:
+            if relation.app:
+                remote_app_databag = relation.data[relation.app]
```
PietroPasotti, line 354: this used to raise ModelError/KeyError at some point, be careful

src/cosl/interfaces/cluster.py
```diff
@@ -341,6 +346,24 @@ def get_address_from_role(self, role: str) -> Optional[str]:
             return address_set.pop()
         return None
 
+    def gather_workload_versions(self) -> Set[str]:
+        """Gather workers' published workload versions."""
+        data: Set[str] = set()
```
PietroPasotti, line 351:
```suggestion
        workload_versions: Set[str] = set()
```

src/cosl/interfaces/cluster.py
```diff
@@ -453,14 +479,17 @@ def publish_unit_address(self, url: str):
             unit_databag = relation.data[self.model.unit]  # type: ignore # all checks are done in __init__
             databag_model.dump(unit_databag)
 
-    def publish_app_roles(self, roles: Iterable[str]):
-        """Publish this application's roles via the application databag."""
+    def publish_app_data(self, roles: Iterable[str], version: Optional[str]):
```
PietroPasotti, line 482:
```suggestion
    def publish_app_data(self, roles: Iterable[str], workload_version: Optional[str]):
```

src/cosl/interfaces/cluster.py
```diff
@@ -479,6 +508,13 @@ def _get_data_from_coordinator(self) -> Optional[ClusterProviderAppData]:
 
         return data
 
+    def get_worker_config_version(self) -> Optional[str]:
+        """Fetch the worker config workload version from the coordinator databag."""
```
PietroPasotti, line 512:
```suggestion
        """Fetch the worker config version from the coordinator databag."""
```

Or, if you go with the suggestion above:
```suggestion
    def is_config_valid_for_version(self) -> bool:
        """Check whether the config we received is valid for the version we requested."""
                data = self._get_data_from_coordinator()
        if data:
            return data.workload_version_supported
        return False
        ```

PietroPasotti (changes requested):
I think we should simplify some things.
GJ with the testing suite!

sed-i:
Hey @michaeldmitry 
I don't quite understand what's going on here.
- Seems like you're introducing a new conceptual idea to charming, but it seems to be too coupled to the coordinator code. Would it be possible to somehow extract it so it's less tightly coupled?
- Would you be able to post a writeup on the core ideas on discourse?

src/cosl/coordinated_workers/coordinator.py
michaeldmitry: Ok I see your point :D It's not that solid use case to use the factory pattern. I like your approach of splitting the legacy and the modern config generation classes.

PietroPasotti:
> Hey @michaeldmitry I don't quite understand what's going on here.
> 
> * Seems like you're introducing a new conceptual idea to charming, but it seems to be too coupled to the coordinator code. Would it be possible to somehow extract it so it's less tightly coupled?
> * Would you be able to post a writeup on the core ideas on discourse?

updated the PR description a little, hope this helps.
This is not a 'new conceptual idea' I think, rather a coordinator-specific feature, hence the coupling.

michaeldmitry:
We'll not be implementing this now.
Will revisit later.

PR #131, opened 2025-02-28
Feat/open specific worker ports

PietroPasotti:
## Issue
This PR adds to the cluster schema a field that lets the coordinator charm decide what ports the worker charm will open.

Before: 
the worker charm was forced to open all possible ports it might have
Now:
the coordinator has the single source of truth of what servers will listen on which ports (depending on the worker role) and assigns to each worker the ports it should be listening on.

## Testing Instructions
Tandem PRs: 
- https://github.com/canonical/charm-relation-interfaces/pull/224
- https://github.com/canonical/tempo-coordinator-k8s-operator/pull/123 
- https://github.com/canonical/tempo-worker-k8s-operator/pull/63

All changes are backwards-compatible: the field is optional, it's up to the worker to gracefully fall back to the old behaviour if the field is missing or null.

# Upgrade notes for coordinated-workers charms

**Coordinator**: 
- pass to the `Coordinator` object a `worker_ports` callable that takes a worker role str as input and returns a tuple of integers (ports) that should be opened by the worker units with that role.
**Worker**:
- Replace `self.unit.open_ports(*LEGACY_PORTS)`  with something like:

> self.unit.open_ports(*(Worker.cluster.get_worker_ports() or LEGACY_PORTS))

Example for mimir:
- https://github.com/canonical/mimir-worker-k8s-operator/pull/90
- https://github.com/canonical/mimir-coordinator-k8s-operator/pull/113


lucabello (approved):
LGTM, but I have one extra ask (if possible)!

Mimir and Loki easily fall behind on cos-lib versions, and since most of them require changes in the main charm (not that it's a breaking change, but introduces things we **should** use when we update), it's very hard to piece together how to update them.

I'm wondering if we could have, in PR descriptions for cos-lib, "upgrade instructions" for a *coordinated-workers* charm, so that when we have time to update Mimir/Loki, we can open the cos-lib releases page and follow instructions for each release.

PR #143, opened 2025-05-01
Lift coordinators' nginx config generation to cosl 

michaeldmitry:
## Issue
All our (HA) coordinators generate their own nginx config from scratch although they share a lot of common config.

## Solution
Extract all shared configuration into `cosl`, and keep only the coordinator-specific parts of the `upstreams` and `locations` blocks, which each coordinator will pass to the shared object as needed.

## Context
- We do nginx config generation using `crossplane`, so we lazy load it inside `NginxConfig` to avoid adding a new dependency to `cosl`

## Testing Instructions
Tandem PRs:
- https://github.com/canonical/tempo-coordinator-k8s-operator/pull/155
- https://github.com/canonical/loki-coordinator-k8s-operator/pull/52
- https://github.com/canonical/parca-k8s-operator/pull/444
- https://github.com/canonical/mimir-coordinator-k8s-operator/pull/130

## Upgrade Notes
- This introduces a breaking change to the coordinated workers. The coordinator used to accept `nginx_config: Callable[["Coordinator"], str]`, but in order to avoid getting this messy with custom conditions/checks, we can just replace it with a different expected input. (We can make it backwards-compatible if we need to, but at some point we'd need to have a breaking change anyways :D)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "cosl"
-version = "0.0.57"
+version = "0.1.0"
 authors = [
     { name = "sed-i", email = "82407168+sed-i@users.noreply.github.com" },
 ]
@@ -104,4 +104,3 @@ ignore-words-list = "assertIn"
 [tool.hatch.metadata]
 # allow git+ dependencies in pyproject
 allow-direct-references = true
-
--- a/src/cosl/coordinated_workers/nginx.py
+++ b/src/cosl/coordinated_workers/nginx.py
@@ -124,9 +124,9 @@ class NginxConfig:
         """
         try:
             # we lazy-load crossplane to avoid adding a dependency to cosl as a whole
-            import crossplane
+            import crossplane  # type: ignore
 
-            self._builder = crossplane.build
+            self._builder = crossplane.build  # type: ignore
         except ModuleNotFoundError:
             raise RuntimeError(
                 "Unmet dependency: the coordinated-workers NginxConfig object depends on the 'crossplane' package. \
@@ -143,7 +143,7 @@ class NginxConfig:
     def get_config(self, addresses_by_role: Dict[str, Set[str]], tls: bool) -> str:
         """Render the Nginx configuration as a string."""
         full_config = self._prepare_config(addresses_by_role, tls)
-        return self._builder(full_config)
+        return self._builder(full_config)  # type: ignore
 
     def _prepare_config(
         self, addresses_by_role: Dict[str, Set[str]], tls: bool
@@ -269,8 +269,10 @@ class NginxConfig:
 
         return nginx_upstreams
 
-    def _build_servers_config(self, backends: List[str], tls: bool = False) -> List[Any]:
-        servers = []
+    def _build_servers_config(
+        self, backends: List[str], tls: bool = False
+    ) -> List[Dict[str, Any]]:
+        servers: List[Dict[str, Any]] = []
         for port, locations in self._server_ports_to_locations.items():
             server_config = self._build_server_config(
                 port,
@@ -291,9 +293,9 @@ class NginxConfig:
     ) -> Dict[str, Any]:
         auth_enabled = False
         is_grpc = any(loc.is_grpc for loc in locations)
-        locations = self._locations(locations, is_grpc, backends, tls)
+        nginx_locations = self._locations(locations, is_grpc, backends, tls)
         server_config = {}
-        if len(locations) > 0:
+        if len(nginx_locations) > 0:
             server_config = {
                 "directive": "server",
                 "args": [],
@@ -321,7 +323,7 @@ class NginxConfig:
                         if tls
                         else []
                     ),
-                    *locations,
+                    *nginx_locations,
                 ],
             }
 
@@ -336,7 +338,7 @@ class NginxConfig:
     ) -> List[Dict[str, Any]]:
         s = "s" if tls else ""
         protocol = f"grpc{s}" if grpc else f"http{s}"
-        nginx_locations = []
+        nginx_locations: List[Dict[str, Any]] = []
 
         if self._enable_health_check:
             nginx_locations.append(
@@ -424,7 +426,7 @@ class NginxConfig:
         return []
 
     def _listen(self, port: int, ssl: bool, http2: bool) -> List[Dict[str, Any]]:
-        directives = []
+        directives: List[Dict[str, Any]] = []
         directives.append(
             {"directive": "listen", "args": self._listen_args(port, False, ssl, http2)}
         )
@@ -438,7 +440,7 @@ class NginxConfig:
         return directives
 
     def _listen_args(self, port: int, ipv6: bool, ssl: bool, http2: bool) -> List[str]:
-        args = []
+        args: List[str] = []
         if ipv6:
             args.append(f"[::]:{port}")
         else:
@@ -552,7 +554,7 @@ class Nginx:
     def configure_pebble_layer(self) -> None:
         """Configure pebble layer."""
         if self._container.can_connect():
-            new_config: str = self._config_getter(tls=self.are_certificates_on_disk)
+            new_config: str = self._config_getter(self.are_certificates_on_disk)
             should_restart: bool = self._has_config_changed(new_config)
             self._container.push(self.config_path, new_config, make_dirs=True)  # type: ignore
             self._container.add_layer("nginx", self.layer, combine=True)
--- a/src/cosl/coordinated_workers/nginx.py
+++ b/src/cosl/coordinated_workers/nginx.py
@@ -5,7 +5,7 @@
 
 import logging
 import subprocess
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Set, TypedDict
@@ -50,10 +50,30 @@ class NginxLocationModifier(Enum):
 class NginxLocationConfig:
     """Represents a `location` block in an Nginx configuration."""
 
-    upstream: str
     path: str
+    backend: str
+    backend_url: str = ""
+    headers: Dict[str, str] = field(default_factory=dict)
     modifier: NginxLocationModifier = NginxLocationModifier.PREFIX
     is_grpc: bool = False
+
+
+@dataclass
+class NginxUpstream:
+    """Represents metadata needed to construct an Nginx `upstream` block."""
+
+    name: str
+    """Name of the upstream block."""
+    port: int
+    """Port number that all backend servers in this upstream listen on.
+
+    Our coordinators assume that all servers under an upstream share the same port.
+    """
+    worker_role: str
+    """The worker role that corresponds to this upstream.
+
+    This role will be used to look up workers (backend server) addresses for this upstream.
+    """
 
 
 class NginxConfig:
@@ -62,9 +82,36 @@ class NginxConfig:
     def __init__(
         self,
         server_name: str,
-        roles_to_upstreams: Dict[str, Dict[str, int]],
+        upstream_configs: List[NginxUpstream],
         server_ports_to_locations: Dict[int, List[NginxLocationConfig]],
+        enable_health_check: bool = False,
+        enable_status_page: bool = False,
     ):
+        """Constructor for an Nginx config generator object.
+
+        Args:
+            server_name: The name of the server (e.g. coordinator fqdn), which is used to identify the server in Nginx configurations.
+            upstream_configs: List of Nginx upstream metadata configurations used to generate Nginx `upstream` blocks.
+            server_ports_to_locations: Mapping from server ports to a list of Nginx location configurations.
+            enable_health_check: If True, adds a `/` location that returns a basic 200 OK response.
+            enable_status_page: If True, adds a `/status` location that enables `stub_status` for basic Nginx metrics.
+
+        Example:
+            .. code-block:: python
+            NginxConfig(
+            server_name = "tempo-coordinator-0.tempo-coordinator-endpoints.model.svc.cluster.local",
+            upstreams = [
+                NginxUpstream(name="zipkin", port=9411, worker_role="distributor"),
+            ],
+            server_ports_to_locations = {
+                9411: [
+                    NginxLocationConfig(
+                        path="/",
+                        backend="zipkin"
+                    )
+                ]
+            })
+        """
         try:
             # we lazy-load crossplane to avoid adding a dependency to cosl as a whole
             import crossplane
@@ -78,8 +125,10 @@ class NginxConfig:
         self._server_name = server_name
         self._dns_IP_address = self._get_dns_ip_address()
         self._ipv6_enabled = is_ipv6_enabled()
-        self._roles_to_upstreams = roles_to_upstreams
+        self._upstream_configs = upstream_configs
         self._server_ports_to_locations = server_ports_to_locations
+        self._enable_health_check = enable_health_check
+        self._enable_status_page = enable_status_page
 
     def get_config(self, addresses_by_role: Dict[str, Set[str]], tls: bool) -> str:
         """Render the Nginx configuration as a string."""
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "cosl"
-version = "0.0.57"
+version = "0.1.0"
```
PietroPasotti, line 7: :birthday:

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
     "nginx_port": 8080,
     "nginx_exporter_port": 9113,
 }
+RESOLV_CONF_PATH = "/etc/resolv.conf"
+
+
+class NginxLocationModifier(Enum):
+    """Enum representing valid Nginx `location` block modifiers."""
+
```
PietroPasotti, line 39:
```suggestion
# cfr. https://www.digitalocean.com/community/tutorials/nginx-location-directive#nginx-location-directive-syntax

```

PietroPasotti: I don't see why we shouldn't add all valid options while we're at it

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+    EXACT = "="  # exact match
+    REGEX = "~"  # case-sensitive regex
+
+
+@dataclass
+class NginxLocationConfig:
+    """Represents a `location` block in an Nginx configuration."""
+
+    upstream: str
+    path: str = "/"
```
PietroPasotti, line 50: I'd make path mandatory (i.e. no default). it's good to force the user to be explicit about the most crucial arguments.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+    modifier: NginxLocationModifier = NginxLocationModifier.DEFAULT
+    is_grpc: bool = False
+
+
+class NginxConfig:
+    """Builds the Nginx configuration."""
```
PietroPasotti, line 56:
```suggestion
    """Responsible for building the Nginx configuration used by the Workers."""
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+
+    def __init__(
+        self,
+        server_name: str,
+        roles_to_upstreams: Dict[str, Dict[str, int]],
```
PietroPasotti, line 61: I'd add a docstring with some clarifications about the content of these vars and an example. These data structures are quite packed with info.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+        server_ports_to_locations: Dict[int, List[NginxLocationConfig]],
+    ):
+        try:
+            # TODO: lazy load crossplane to avoid adding an additional dependency to cosl
```
PietroPasotti, line 65:
```suggestion
            # we lazy-load crossplane to avoid adding a dependency to cosl as a whole
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+            import crossplane
+
+            self._builder = crossplane.build
+        except ModuleNotFoundError:
+            raise RuntimeError(
+                "Unmet dependencies: the coordinator charm base is missing some dependencies. \
```
PietroPasotti, line 71:
```suggestion
                "Unmet dependency: the coordinated-workers NginxConfig object depends on the 'crossplane' package. \
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+            Please install them with: pip install crossplane"
```
PietroPasotti, line 72:
```suggestion
            Please install it with: `pip install crossplane` and add it to your charm's dependencies."
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+            )
+        self._server_name = server_name
+        self._dns_IP_address = self._get_dns_ip_address()
+        self._ipv6_enabled = is_ipv6_enabled()
+        self._roles_to_upstreams = roles_to_upstreams
+        self._server_ports_to_locations = server_ports_to_locations
+
+    def get_config(self, addresses_by_role: Dict[str, Set[str]], tls: bool) -> str:
+        """Returns the rendered Nginx configuration as a string."""
```
PietroPasotti, line 81:
```suggestion
        """Render the Nginx configuration as a string."""
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,319 @@
...
+        full_config = self._prepare_config(addresses_by_role, tls)
+        return self._builder(full_config)
+
+    def _prepare_config(
+        self, addresses_by_role: Dict[str, Set[str]], tls: bool
+    ) -> List[Dict[str, Any]]:
+        log_level = "error"
```
PietroPasotti, line 88: we're using this var only once? let's nuke it

PietroPasotti (changes requested):
Overall looks good. I think we should add better docstrings and especially a couple of tests to validate the generated config.

I would: grab a couple of nginx.conf files from actual deployments (with/without ingress, with/without tls) for tempo, loki and mimir, and save those files in a directory in the tests folder, then have unittests tests that do:

```
def test_validate_nginx_config_generation():
    generated_conf = NginxConfig(...).get_config()  # << figuring out these args is going to be the hard part
    assert Path(./resources/sample_tempo_conf_no_tls.txt).read_text() == generated_conf
```

I don't think we should aim for a 100% correct match, so it's fine if we go and modify the stored sample files to match what is being generated, but we should do this as an exercise to double-check that we're not omitting crucial parts or making unwanted semantic changes.

src/cosl/coordinated_workers/nginx.py
michaeldmitry: Added them all

tests/test_coordinated_workers/test_nginx.py
```diff
@@ -139,3 +148,146 @@ def test_has_config_changed(nginx_context: testing.Context):
...
+                path="/loki/api/v1/status/buildinfo", modifier="=", backend="worker"
+            ),
+            NginxLocationConfig(path="/ring", modifier="=", backend="worker"),
+        ]
+    },
+}
+
+
+def _get_nginx_config_params(workload: str) -> Tuple[list, dict]:
+    return (upstream_configs.get(workload), server_ports_to_locations.get(workload))
```
PietroPasotti, line 293:
```suggestion
    return upstream_configs[workload], server_ports_to_locations[workload]
```

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,413 @@
...
+    EXACT = "="  # exact match
+    REGEX_CASE_SENSITIVE = "~"  # case-sensitive regex match
+    REGEX_CASE_INSENSITIVE = "~*"  # case-insensitive regex match
+    PREFIX_NO_REGEX = "^~"  # prefix match that disables further regex matching
+
+
+@dataclass
+class NginxLocationConfig:
+    """Represents a `location` block in an Nginx configuration."""
+
```
PietroPasotti, line 52: perhaps add a comment/docstring code block that shows what it means.
E.g.

```suggestion


# for example, NginxLocationConfig('/', 'foo.com', headers={'a': 'b'}, modifier=EXACT, is_grpc=True) 
# would result in:
# ... nginx.conf
#  / = ...
```

PietroPasotti (approved):
nice! Let's get the chip in from the core team, then we can proceed.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,425 @@
...
+    """Represents a `location` block in an Nginx configuration.
+
+    For example, NginxLocationConfig('/', 'foo', backend_url="/api/v1" headers={'a': 'b'}, modifier=EXACT, is_grpc=True)
+    would result in:
+        location = / {
+            set $backend grpc://foo/api/v1;
+            grpc_pass $backend;
+            proxy_connect_timeout 5s;
+            proxy_set_header a b;
+        }
```
Abuelodelanada, line 60: Shouldn't we have a docstring in this file that explains the usage of the objects provided by this lib?

For instance, we have that in our charm libs: https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v1/loki_push_api.py#L7

MichaelThamm:
Due to a somewhat urgent need from Field to release Loki and Mimir fixes for dynamic Nginx resolver config from `/etc/resolv.conf`, we have related PRs:
- https://github.com/canonical/mimir-coordinator-k8s-operator/pull
- loki ... (when its created)

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,425 @@
...
+        try:
+            # we lazy-load crossplane to avoid adding a dependency to cosl as a whole
+            import crossplane  # type: ignore
+
+            self._builder = crossplane.build  # type: ignore
+        except ModuleNotFoundError:
+            raise RuntimeError(
+                "Unmet dependency: the coordinated-workers NginxConfig object depends on the 'crossplane' package. \
+            Please install it with: `pip install crossplane` and add it to your charm's dependencies."
+            )
```
Abuelodelanada, line 134: Are we sure this is a good idea?

We created cos-lib among other things to be able to avoid these kinds of things that are common in charm libs where we are not able to add external dependencies.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,425 @@
...
+                    {"directive": "scgi_temp_path", "args": ["/tmp/scgi_temp"]},
+                    # logging
+                    {"directive": "default_type", "args": ["application/octet-stream"]},
+                    {
+                        "directive": "log_format",
+                        "args": [
+                            "main",
+                            '$remote_addr - $remote_user [$time_local]  $status "$request" $body_bytes_sent "$http_referer" "$http_user_agent" "$http_x_forwarded_for"',
+                        ],
+                    },
```
Abuelodelanada, line 188: I would not hardcode these values here, I would them as module's constants:

- "/dev/stderr"
- "/tmp/nginx.pid"
- "8192"
- "4096"
...

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +31,425 @@
...
+                        "args": ["X-Scope-OrgID", "$ensured_x_scope_orgid"],
+                    },
+                    {"directive": "server_name", "args": [self._server_name]},
+                    *(
+                        [
+                            {"directive": "ssl_certificate", "args": [CERT_PATH]},
+                            {"directive": "ssl_certificate_key", "args": [KEY_PATH]},
+                            {
+                                "directive": "ssl_protocols",
+                                "args": ["TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"],
```
Abuelodelanada, line 316: From my point of view, these TLS protocol versions are important enough to extract them as modules constants. This way it is much more clear to know about the supported versions at glance.

Abuelodelanada (changes requested):

src/cosl/coordinated_workers/nginx.py
michaeldmitry: We already have `crossplane` as a dependency in our HA charms. If you're talking about the lazy-loading, this is to avoid adding a `cosl`-wide dependency to users.

Abuelodelanada: What I mean is that before this change, we import `crossplane` in every coordinator charm.

For instance, in Loki coordinator we import it [here](https://github.com/canonical/loki-coordinator-k8s-operator/blob/main/src/nginx_config.py#L9), and it is used by the [NginxConfig](https://github.com/canonical/loki-coordinator-k8s-operator/blob/main/src/nginx_config.py#L142)

Now we are moving the `NginxConfig` to cos-lib and [removing it](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/155/files#diff-0116a9b741399a7bac8d1015f203baf6f53391c2ba9b7320fe04876e006c0f62L9) from the coordinators. That's why I think we should import `crossplane` in this module.

From my point of view, if we are not importing `crossplane` in the coordinators any more, we should not add it there as a dependency as well. The dependency should be added in this lib.


Or am I missing something? 🤔

michaeldmitry: Yeah, but adding crossplane to cosl will require every other user who's using cosl (not for the coordinated workers) to have crossplane as a dependency now. Importing it inside NginxConfig helps us limit adding that dep only for users who actually want to use NginxConfig.

Let me know if I'm still missing it and we can discuss it over a call

PietroPasotti: i don't think anyone is missing anything, it's just a tradeoff with either side having its pros and cons. We should make a decision and agree on it.

PietroPasotti: IMHO this approach is the cleaner one for now. In the long term, we should consider moving coordinated-workers out of cosl and into its own library. 
Cosl is becoming a dumping ground for all that is needed by more than 1 charm, but this isn't scalable.

I think avoiding to bloat all charms using cosl with a crossplane dependency they won't need is to be avoided and wins the argument, but I'm happy to change my mind.
@simskij @sed-i quick decision plz

PietroPasotti: https://github.com/canonical/cos-lib/issues/144

Abuelodelanada: > In the long term, we should consider moving coordinated-workers out of cosl and into its own library.

I agree with this, let's move forward.

Abuelodelanada (approved):

michaeldmitry:
> From a review in another Nginx PR, there were some minor improvement comments:
> 
> 1. Use `Syntax:	resolver address ... [valid=time]` to shorten the `valid` attribute on the resolver to something like `5s` as by default it'll use the default DNS TTL which might be large. In the backend for reverse proxy, you usually want to be reactive to change.
> 2. Shorten the default DNS timeout (`30s`, huge when requests are pending to be proxied) with a `resolver_timeout` to something like `5s`.
> 
> To be clear, these are not blocking improvement requests and can be done in another PR or rejected if deemed unnecessary.
@MichaelThamm I don't know if these changes will affect our config negatively or not. So, I think let's open a separate issue for them.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -1,12 +1,164 @@
...
+            )
+        ...
+        @property
+        def hostname(self) -> str:
+            return socket.getfqdn()
+
+        @property
+        def _nginx_locations(self) -> List[NginxLocationConfig]:
+            return [
+                NginxLocationConfig(path="/loki/api/v1/rules", backend="backend",modifier=NginxLocationModifier("=")),
```
PietroPasotti, line 60: this is how I'd use an enum
```suggestion
                NginxLocationConfig(path="/loki/api/v1/rules", backend="backend",modifier=NginxLocationModifier.EXACT),
```

Which makes me think, if it's better to use Typing.Literal instead of this humongously-named enum if all we're really doing it for is type hinting the input.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -29,6 +181,478 @@
...
+            raise RuntimeError(
+                "Unmet dependency: the coordinated-workers NginxConfig object depends on the 'crossplane' package. \
+            Please install it with: `pip install crossplane` and add it to your charm's dependencies."
+            )
+        self._server_name = server_name
+        self._dns_IP_address = self._get_dns_ip_address()
+        self._ipv6_enabled = is_ipv6_enabled()
+        self._upstream_configs = upstream_configs
+        self._server_ports_to_locations = server_ports_to_locations
+        self._enable_health_check = enable_health_check
```
PietroPasotti, line 314: very minor nit, but I tend to try and do all 'init-arg-to-attr' assignments at the top of init, before any method calls, function calls, or other 'derived' attribute initialization. that way it's clearer to read where the init arguments are used.

src/cosl/coordinated_workers/nginx.py
```diff
@@ -133,7 +757,7 @@ def reload(self) -> None:
     def configure_pebble_layer(self) -> None:
         """Configure pebble layer."""
         if self._container.can_connect():
-            new_config: str = self._config_getter()
+            new_config: str = self._config_getter(self.are_certificates_on_disk)
```
PietroPasotti, line 760:
```suggestion
            new_config: str = self._config_getter(tls=self.are_certificates_on_disk)
```
more explicit what it's doing

PietroPasotti (approved):
Looks good!

PR #153, opened 2025-09-04
feat: add statuses based on free disk space

### LINKED ISSUE #152: Add a disk space check that other charms can import
Currently, when other observability charms are deployed without overlays, they get a storage of size 1GiB. There are plans to modify the `charmcraft.yaml` files of our charms so they enforce a higher storage when being deployed. In the meantime, we want the charms to be blocked unless they're deployed using an overlay specifying a storage of size > 1GiB. This would have to be achieved by a check on `pebble_ready`.
In addition, we want the check to be run during `update_status` hooks so that the charm is blocked if storage drops too low - that is, < 1GiB. 

There is a [PR for Prometheus](https://github.com/canonical/prometheus-k8s-operator/pull/719) which applies this logic. We need this logic to be added in other charms such as Loki Monolithic, Loki worker, Tempo worker, etc.

To avoid code duplication across multiple repos, a potential solution would be to introduce a `check_disk_space` logic in `cos-lib`, which said charms can import. As stated before, the check would have to be run on the `pebble_ready` and `update_status` hooks.

sinapah:
## Issue

This PR fixes #152.

## Solution

1. Adds function which accepts the location of storage and the threshold and returns an appropriate status based on the comparison.
2. Adds unit tests for this logic.

## Context

Please review the attached issue (#152) on why this change is being implemented. The idea is that other charms would import this functionality from `cosl` and use it `on_collect_unit_status` to check whether to set status blocked based on whether there is sufficient space remaining or not. The charm has to be blocked if remaining storage is < 1GiB.
## Testing Instructions

The unit test added must pass (and it does).

### DIFF
--- a/src/cosl/statuses.py
+++ b/src/cosl/statuses.py
@@ -17,16 +17,20 @@ def get_disk_usage_status(location: str, *, threshold: int = 1024**3) -> ops.Sta
     """Returns a status that matches the disk usage.
 
     Returns:
-     - ActiveStatus when the provided <location> has more <threshold> bytes;
+     - ActiveStatus when the provided <location> has more than <threshold> free bytes;
      - BlockedStatus when <location> is below <threshold>;
      - MaintenanceStatus when the <location> is not found i.e. before storage is attached.
+
+    Args:
+        location (str): The file system path to check for available disk space e.g. /var/lib/prometheus. In most cases, the charm calling this function would get this by calling self.model.storages["database"][0].location.
+        threshold (int): The minimum number of free bytes required before blocking. The default is 1024**3.
     """
     try:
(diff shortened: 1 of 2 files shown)


src/cosl/statuses.py
```diff
@@ -0,0 +1,38 @@
...
+import shutil
+
+import ops
+
+logger = logging.getLogger(__name__)
+
+
+def get_disk_usage_status(location: str, *, threshold: int = 1024**3) -> ops.StatusBase:
+    """Returns a status that matches the disk usage.
+
```
Abuelodelanada, line 18: Let's add an "Args" section to the docstring to better explain what this function receives.

PR #155, opened 2025-09-26
feat: Cosl owns `type_convert_stored`

MichaelThamm:
## Issue

Many of our libs own the `_type_convert_stored` method:
- graf_dashboard
- graf_datasource
- prom_scrape
- nrpe_exporter
- graf_auth
- ingress_per_unit

This is code duplication.

## Solution

Cosl now owns this and should be imported from in all our libs.

We know that this function works because it is duplicated in all our libraries and I copied it over. The only difference being additional type checking to pass `cosl` static checks and making it public:

image

### DIFF
--- a/src/cosl/types.py
+++ b/src/cosl/types.py
@@ -1,8 +1,10 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 """Types used by cos-lib."""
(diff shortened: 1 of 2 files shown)


Abuelodelanada (approved):
It is a great addition to the lib!

PR #163, opened 2025-10-31
fix: change DashboardPath40UID.generate signature

Abuelodelanada:
## Issue

This PR solves this issue: https://github.com/canonical/grafana-k8s-operator/issues/458

## Solution

The signature of the method `generate` in `DashboardPath40UID` was modified, so it can receive an arbitrary number of arguments, and by combining those it'll generate an UID.

## Context

> Currently, we generate a dashboard''s UID from _charm_ name + dashboard path.
This is correct for all charms except cos-config: We could have multiple cos-config apps in the same deployment, with dashboards such as `./dashboards/overview.json`, in both cases, which would result in uid collision.

## Testing Instructions

Run unit tests ;-)

```
tox -e unit
```

### DIFF
--- a/tests/test_grafana_dashboard.py
+++ b/tests/test_grafana_dashboard.py
@@ -106,5 +106,5 @@ class TestGenerateUID(unittest.TestCase):
             DashboardPath40UID.is_valid("a1b2c3d4e5f6789012345678901234567890abcd")
         )  # 40 chars
         # Generated UIDs should always be valid (covered by other tests but explicitly stated here)
-        sample_uid = DashboardPath40UID.generate("test", "sample")
+        sample_uid = DashboardPath40UID.generate("some-charm", "dashboard.json")
         self.assertTrue(DashboardPath40UID.is_valid(sample_uid))


src/cosl/grafana_dashboard.py
```diff
@@ -75,10 +75,10 @@ def _hash(cls, components: Tuple[str, ...], length: int) -> str:
...
 
     @classmethod
-    def generate(cls, charm_name: str, dashboard_path: str) -> str:
-        """Generate a dashboard uid from charm name and dashboard path.
+    def generate(cls, *args: str) -> str:
+        """Generate a dashboard uid from any number of string components.
 
-        The combination of charm name and dashboard path (relative to the charm root) is guaranteed to be unique across
+        The combination of all provided string components is guaranteed to be unique across
         the ecosystem. By design, this intentionally does not take into account instances of the same charm with
```
MichaelThamm:
```suggestion
        The combination of all provided args must guarantee uniqueness across
        the ecosystem. In general, charm name and dashboard path (relative to the charm root) is guaranteed to be unique across the ecosystem. By design, this intentionally does not take into account instances of the same charm with
```

PR #169, opened 2025-11-21
feat: add `aggregator metrics missing` rule

### LINKED ISSUE #165: HostMetricsMissing only alerts when ALL units are down
### Bug Description

When a subset of aggregator units are down, the `absent(up)` metric does not trigger an alert unless all units are down.

> [!IMPORTANT]
> This needs to be backported into `1/stable`

We need to add a test for this case in cos-lib for scaled aggregators with a unit going down!

### To Reproduce

1. Deploy in a k8s model:
```yaml
bundle: kubernetes
saas:
  remote-f4348ce868694e0e895b8a123f861363: {}
applications:
  alertmanager:
    charm: alertmanager-k8s
    channel: 2/edge
    revision: 191
    resources:
      alertmanager-image: 101
    scale: 1
    constraints: arch=amd64
    storage:
      data: kubernetes,1,1024M
    trust: true
  prom:
    charm: prometheus-k8s
    channel: 2/edge
    revision: 272
    resources:
      prometheus-image: 152
    scale: 1
    constraints: arch=amd64
    storage:
      database: kubernetes,1,1024M
    trust: true
  traefik:
    charm: traefik-k8s
    channel: latest/stable
    revision: 223
    base: ubuntu@20.04/stable
    resources:
      traefik-image: 162
    scale: 1
    constraints: arch=amd64
    storage:
      configurations: kubernetes,1,1024M
    trust: true
relations:
- - traefik:ingress-per-unit
  - prom:ingress
- - traefik:ingress
  - alertmanager:ingress
- - prom:alertmanager
  - alertmanager:alerting
- - prom:metrics-endpoint
  - traefik:metrics-endpoint
- - prom:receive-remote-write
  - remote-f4348ce868694e0e895b8a123f861363:send-remote-write
--- # overlay.yaml
applications:
  prom:
    offers:
      prom:
        endpoints:
        - receive-remote-write
        acl:
          admin: admin
```
2. Deploy in a LXD model:
```yaml
default-base: ubuntu@22.04/stable
saas:
  prom:
    url: microk8s:admin/man-sol.prom
applications:
  gagent:
    charm: grafana-agent
    channel: 1/stable
    revision: 603
    trust: true
  zoo:
    charm: zookeeper
    channel: 3/stable
    revision: 158
    num_units: 2
    to:
    - "0"
    - "1"
    constraints: arch=amd64
    storage:
      data: rootfs,1,1024M
    trust: true
machines:
  "0":
    constraints: arch=amd64
  "1":
    constraints: arch=amd64
relations:
- - zoo:cos-agent
  - gagent:cos-agent
- - gagent:send-remote-write
  - prom:receive-remote-write
```
3. `jssh gagent/0 "sudo snap stop grafana-agent"`
4. Check in Prometheus for the Zookeeper `_gagent_AggregatorHostHealth_alerts::HostMetricsMissing` alert which is not firing
5. Run the query manually:
    - `absent(up{juju_application="gagent",juju_model="foo",juju_model_uuid="foo", juju_unit="gagent/0"})` -> correctly shows host down
    - `absent(up{juju_application="gagent",juju_model="foo",juju_model_uuid="foo", juju_unit="gagent/1"})` -> correctly shows host up
    - `absent(up{juju_application="gagent",juju_model="foo",juju_model_uuid="foo"})` -> shows the application is still available (but at least 1 unit is down)

### Environment

.

### Relevant log output

```shell
.
```

### Additional context

_No response_

### LINKED ISSUE #168: Create another generic alert rule for HostMetricsMissing of severity critical
**Feature tracking issue (currently draft)**

We currently have the HostMetricsMissing generic aggregator rule which is supposed to be triggered when `up` has been missing from an aggregator (e.g. a remote writer) for 5 minutes. We have an issue #165 which outlines that this rule only fires when all units of the aggregator are down. This leads to blind spots, as we need an alert even when only one of multiple units is down.

Our initial approach in https://github.com/canonical/prometheus-k8s-operator/pull/748 was to duplicate the HostMetricsMissing alert per unit of a remote writer. A small issue with this approach is that the alert for each unit would then have severity=critical, which would go against HA properties. What we need is warning level alert rules per unit and a single critical alert when all units are down (essentially what we already have for HostMetricsMissing).

To summarize:
1. We would keep HostMetricsMissing but change the annotations to explicitly clarify that this alert is a sign that a single unit of a remote writer is down/failing to remote write. This alert would be duplicated per unit, which is the approach 748 takes. There would be a need to inject `juju_unit` into the alert expression or alert labels. For K8s charms, this rule would have severity=warning. For machine charms, the severity would be critical. The remote writer itself (based on peer relations) should be able to tell how many units it has and should then duplicate this rule (to be defined in cosl.generic_alert_rules) per each unit and inject the corresponding unit number into the alert label and expression.
2. We would add another alert rule, similar to the one above but for the app level e.g. AggregatorMetricsMissing. This rule does not need to have the juju unit in the alert labels or expression, since it's app level. It should be triggered when all units are down and always have severity critical.

To achieve this, we have the following two action items:
1. Update `cosl` so we have the unit level alert defined there.
2. Update `PrometheusRemoteWriteConsumer` in the Prometheus repo so it duplicates the `HostMetricsMissing` alert rule per each unit of a remote writer and injects `juju_unit` into the alert expression and alert labels using CosTool.

sinapah:
## Issue

Fixes #168 and partially fixes #165.

## Solution

1. Adds a new aggregator rule called `AggregatorMetricsMissing`. This will have severity=critical and will fire when all units of a remote writer are down or failing to remote write.
2. Modifies the existing rule called `HostMetricsMissing` so that it will be unit level alert. This means that it will be dedicated to a single unit of an app and fires when that specific unit is down or failing to remote write. It will have severity warning for K8s units and critical for machine units.
After this PR goes through, https://github.com/canonical/prometheus-k8s-operator/pull/748 will need to be modified so it checks charm type and appropriately sets the severity level in (2).

### DIFF
--- a/src/cosl/rules.py
+++ b/src/cosl/rules.py
@@ -94,7 +94,7 @@ from .types import (
 
 logger = logging.getLogger(__name__)
 
-AGGREGATOR_METRICS_MISSING_RULE_NAME = "AggregatorMetricsMissing"
+HOST_METRICS_MISSING_RULE_NAME = "HostMetricsMissing"
 
 _generic_alert_rules: Final = SimpleNamespace(
     # We use "5m" to avoid false positives on expected temporary "down", e.g. during intentional (re)start.
@@ -110,7 +110,7 @@ _generic_alert_rules: Final = SimpleNamespace(
         },
     },
     host_metrics_missing={
-        "alert": "HostMetricsMissing",
+        "alert": HOST_METRICS_MISSING_RULE_NAME,
         "expr": "absent(up)",
         "for": "5m",
         "labels": {
@@ -122,7 +122,7 @@ _generic_alert_rules: Final = SimpleNamespace(
         },
     },
     aggregator_metrics_missing={
-        "alert": AGGREGATOR_METRICS_MISSING_RULE_NAME,
+        "alert": "AggregatorMetricsMissing",
         "expr": "absent(up)",
         "for": "5m",
         "labels": {"severity": "critical"},


src/cosl/rules.py
```diff
@@ -110,14 +110,24 @@
...
     host_metrics_missing={
         "alert": "HostMetricsMissing",
-        # We use "absent(up)" with "for: 5m" because the alert transitions from "Pending" to "Firing".
-        # If query portability is desired, "absent_over_time(up[5m])" is an alternative.
         "expr": "absent(up)",
         "for": "5m",
-        "labels": {"severity": "critical"},
+        "labels": {
+            "severity": "critical"
+        },  # The remote writer will set this to critical for machine charms when initializing PrometheusRemoteWriteConsumer.
```
sed-i: Is this comment relevant here? The severity is already critical.
Iirc we said we'd replicate HostMetricsMissing (this) alert?
- Host = Unit
- Aggregator = App

src/cosl/rules.py
```diff
@@ -110,14 +110,24 @@
...
+        "annotations": {
+            "summary": "Unit '{{ $labels.juju_unit }}' of application '{{ $labels.juju_application }}' is down or failing to remote write.",
+            "description": "`Up` missing for unit '{{ $labels.juju_unit }}' of application {{ $labels.juju_application }} in model {{ $labels.juju_model }}. Please ensure the unit or the collector scraping it is up and is able to successfully reach the metrics backend.",
+        },
+    },
+    aggregator_metrics_missing={
+        "alert": "AggregatorMetricsMissing",
+        "expr": "absent(up)",
+        "for": "5m",
+        "labels": {"severity": "warning"},
```
sed-i: When all units are down, we can it to be critical.

Issue #191, opened 2026-04-03
Redesigning Rules class to support easy addition of new rule types


swetha1654:
I was looking to add support for Sigma rules in the `otlp` interface and noticed that 
it uses the `cosl` library, especially the `Rules` class for rules normalization to a uniform 
format, juju topology injection, and validation. Right now, the `Rules` class is tightly coupled to
Prometheus and Loki and addition of Sigma rules is not straight-forward.

## Analysis of the rule format differences

Before I discuss how to restructure the `Rules` class, here are the key structural differences 
between Prometheus / Loki and Sigma rule formats:

| Aspect   | Prometheus / Loki | Sigma |
|----------|----------|----------|
| Grouping | Rules must live inside `groups`  | Each rule is independent, no concept of grouping |
| Identify  | Group name + alert / record name |  UUID  |
| Detection logic | `expr` field that has `PromQL`/ `LogQL` queries  | `detection` block with `selection` and `condition` |
| Data source | Implicit (the scrape target)  | `logsource` field (product / category / field)  |
| Validation | Valdiated using `cos-tool` | Validated using `pySigma` library |
| Classification of Rules | Rules can be either RecordingRule or AlertingRule | There is only one format |

## Analysis of the current state of Rules class

The `Rules` class has the following public interfaces:

1. `add`: Add rules from dict to the existing ruleset.
2. `add_path`: Add rules from a directory.
3. `inject_and_validate_rules`: Inject Juju topology labels and validate rules using CosTool.
4. `as_dict`: Return standard rules file in dict representation (used to dump rules into the databag).

Internally, the class does the following:

1. Ingest rules from multiple sources (either `add_path` / `add`)
2. Normalize the rules to the "official" format (the rules must have the `groups` key)
3. Injects Juju topology in:
   1. The group name (Prepends topology identifier + relative path to the group name, appends `_rules` suffix)
   2. Labels 
   3. Expressions (modifies `expr` to contain the juju topology)
4. Validates the syntax using `CosTool`.

## Questions

1. I noticed that the Juju topology injection happens both when `add`/`add_path` and 
`inject_and_validate_rules` is called. The former is called on the requirer side while the latter is
used by the provider by injecting the same Juju Topology it receives from the requirer via databag. 
Is there a need to do it both times? It seems a bit redundant to me.
2. Do you see any value in combining `add` and `add_path` into a single interface or keep it seperate?

## Proposal 

[...]

PR #192, opened 2026-04-07
feat: Introduce a new GenericRules class to replace Rules class

swetha1654:
## Issue

POC for #191 

## Tandem PRs

1. charmlibs - [feat(otlp): Update otlp to use generic rules class](https://github.com/canonical/charmlibs/pull/437)
2. otel-collector: [test latest otlp interface](https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/246)

## Visualization

The diagram visualizing the changes in this PR can be found in the [issue](https://github.com/canonical/cos-lib/issues/191#issuecomment-4333126477)

## Solution

Some notes about the new proposal:

1. The existing `AlertRules`, `RecordingRules` and `Rules` class are used by several downstream libraries and repos including: `cos-proxy-operator`, `prometheus_scrape`, `prometheus_remote_write`, `grafana-agent-operator`. Rewriting the `Rules` class would break all of them hence I've proposed a `AbstractRules` class.
4. The `Rules` class implements both the public facing APIs (`add`, `add_path`, `as_dict`) as well the backend specific logic (topology injection, rules validation, group name definition logic). I've broken this down into `AbstractRules` class that exposes the public APIs and a generic `RuleBackend` class that every component uses to implement their own validation and injection logic.
5. `inject_and_validate` from the old `Rules` class has been adopted and modified in `AbstractRules` class to just `validate`. I don't see value in running the injection again on a set of rules that are already injected with juju topology (note that the add method calls `_from_dict` which does the injection. The `add` method is called by the requirer class before dumping the data into a relation databag.)

### Checklist

## Context

This PR makes the Rules class generic enough to accept new Rule types in the future. (for eg: Sigma rules)

### Changes for downstream users

1. The `generic_alert_groups`, `HOST_METRICS_MISSING_RULE_NAME` has been moved to `cosl.prometheus` hence downstream imports of it will fail, but this should be a very easy fix.

There should ideally not be any other breaking changes.


MichaelThamm:
Deprecating AlertRules is certainly a major version bump requirement. We need to be careful because this will break our libs.

swetha1654:
### What I don't like about this design

The usage of kwargs to pass backend specific inputs via the public facing `add` API. Here are the other alternatives I considered -
1. Adding `group_name`/ `group_prefix` as parameters in the `add` API and simply ignoring it in the SigmaBackend implementation.
     Problem: This leads to leakage of backend specific inputs to other backends that don't need them.
2. Creating seperate classes called `PrometheusLokiInputs` and `SigmaRuleInputs` to pass the rule-specific information to the `add` API.
     Problem: Users of the GenericRules class will have yet another class to import and define before they can call the `add` API. The current design is already introducing an extra class for people to have to import. This solution will lead to 3 classes.

Hence I ended up using kwargs. Using kwargs sacrifies type safety but its the standard python pattern for this situation. WDYT?

src/cosl/backends/loki.py
MichaelThamm, line 1: I think, if we do this then we should have a nested module like `cosl.rules.loki`. Loki should not leak into the cos-lib module API. Same comment for `cosl.rules.prometheus`. This does leave the rules.py file in a weird spot though -> `cosl.rules.rules`.

This also applies to `grouped_rules.py`

src/cosl/prometheus.py
```diff
@@ -0,0 +1,122 @@
...
+from .grouped_rules import _GroupedRuleBackend  # type: ignore
+from .types import OfficialRuleFileFormat, QueryType
+
+# ---------------------------------------------------------------------------
+# Generic alert rules (pre-built)
+# ---------------------------------------------------------------------------
+
+HOST_METRICS_MISSING_RULE_NAME = "HostMetricsMissing"
+
+_generic_alert_rules: Final = SimpleNamespace(
```
MichaelThamm: Generic alert rules are not specific to Prometheus and can theoretically include Loki rules or sigma rules in the future

MichaelThamm:
> Any downstream code using the Rules class will break

I thought we said this was not the case? Also, the passing utets indicate that this is not true. Can you update the PR description?

src/cosl/backends/grouped_rules.py
```diff
@@ -0,0 +1,186 @@
...
+            return self.from_dict(
+                rule_file,
+                group_name=file_path.stem,
+                group_name_prefix=group_name_prefix,
+            )
+        except ValueError as e:
+            logger.error("Invalid rules file: %s (%s)", file_path.name, e)
+            return []
+
+    def validate(self, rules: Dict[str, List[OfficialRuleFileItem]]) -> Tuple[bool, str]:
```
MichaelThamm, line 162: `inject_and_validate` -> `validate` would require a tandem PR in the OTLP charmlib. We have [this Jira story ](https://warthogs.atlassian.net/browse/OBC-1838) to remove the method entirely. I will leave this up to you to either:
- Tandem PR in OTLP charmlib
- Keep the `inject_and_validate` since we will remove it eventually anyways

MichaelThamm (changes requested):

MichaelThamm:
```python
from cosl.rules import generic_alert_groups, HOST_METRICS_MISSING_RULE_NAME
# or
from cosl import generic_alert_groups
```
will now fail, and these are used in many places. We need to be careful with backwards compatibility here.

You can add this to `cosl.rules`:
```python
# rules.py — backward-compat re-exports (deprecated, remove in next major)
from .prometheus import generic_alert_groups, HOST_METRICS_MISSING_RULE_NAME  # noqa: F401
```

MichaelThamm (commented):
In the old Rules.add_path:

```Python
else:
    logger.debug("Rules path does not exist: %s", path)
```
In the new GenericRules.add_path:

```Python
else:
    raise InvalidRulePathError(...)
```

I think this difference is fine, because we said that we would allow Rules and GenericRules to exist alongside one another and deprecate in the future.

src/cosl/rules.py
```diff
@@ -208,10 +158,236 @@ def __init__(
...
+    """
+
+    def __init__(self, backend: RuleBackend[T], topology: Optional[JujuTopology] = None):
+        """Build a Rules instance.
+
+        Args:
+            backend: A :class:`RuleBackend` implementation that handles all
+                format-specific logic.
+            topology: An optional :class:`JujuTopology` instance used to
+                annotate all rules.  If provided, it is set on the backend,
```
MichaelThamm:
```suggestion
                annotate all rules. If provided, it is set on the backend,
```

src/cosl/rules.py
```diff
@@ -208,10 +158,236 @@ def __init__(
...
+                overriding any topology already configured there.
+        """
+        self.backend = backend
+        if topology is not None:
+            self.backend.topology = topology
```
MichaelThamm:
```Python
# GenericRules.__init__
if topology is not None:
    self.backend.topology = topology
```
This mutates the backend object passed in by the caller. If a caller reuses a `PrometheusRuleBackend` instance across multiple `GenericRules` instances with different topologies, the last one wins silently. 

If topology is already an attr of the RuleBackend, why even accept a topology param?

swetha1654:
> I thought we said this was not the case? Also, the passing utets indicate that this is not true. Can you update the PR description?

@MichaelThamm my bad i forgot to remove it, I've updated the PR description

src/cosl/backends/grouped_rules.py
swetha1654: The `otlp` library is still using the older `Rules` class which is untouched and still has the `inject_and_validate` method so this wouldn't break the OTLP charmlib. Once this PR is approved and merged I plan on:
1. Adding support for `Sigma` rules backend in `cos-lib` next.
2. Using the newer GenericClass with Sigma backend in `otlp` library.

At step 2, I will migrate the `otlp` library from the old `Rules` class to `GenericRules` and will migrate from `inject_and_validate` to `validate`. Does this sound good @MichaelThamm ?

swetha1654:
> In the old Rules.add_path:
> 
> ```python
> else:
>     logger.debug("Rules path does not exist: %s", path)
> ```
> 
> In the new GenericRules.add_path:
> 
> ```python
> else:
>     raise InvalidRulePathError(...)
> ```
> 
> I think this difference is fine, because we said that we would allow Rules and GenericRules to exist alongside one another and deprecate in the future.

Yes this shouldn't be a breaking change or problem

swetha1654:
> ```python
> from cosl.rules import generic_alert_groups, HOST_METRICS_MISSING_RULE_NAME
> # or
> from cosl import generic_alert_groups
> ```
> 
> will now fail, and these are used in many places. We need to be careful with backwards compatibility here.
> 
> You can add this to `cosl.rules`:
> 
> ```python
> # rules.py — backward-compat re-exports (deprecated, remove in next major)
> from .prometheus import generic_alert_groups, HOST_METRICS_MISSING_RULE_NAME  # noqa: F401
> ```

@MichaelThamm I dont think `from cosl import generic_alert_groups` would even work in the current state since `generic_alert_groups` is not defined in the `__init__.py` file. Now that you mention that `generic_alert_groups` are not specific to Prometheus and can theoretically include Loki rules or sigma rules in the future, I don't know if i want to keep it in the prometheus module anymore so I will move it to `common_rules.py` and import them inside `rules.py` to avoid breaking changes.

src/cosl/prometheus.py
swetha1654: in that case I will move it to a new module `common_rules.py` and import them into `rules.py` for backward compatibility.

src/cosl/backends/loki.py
swetha1654: Moving `rules.py` into `cosl.rules` would be a breaking change. Do we want that?

swetha1654: I'm all in favour of moving common files into shared directory but it would result is failure of imports with `cosl.rules`.

MichaelThamm: Yeah, that is a good point. We probably should move the others though. So we have:
```
.
├── rules.py
├── rules
│   ├── loki.py
│   ├── prometheus.py
```

src/cosl/rules.py
swetha1654: Thats a good catch, initally i didn't have it as a class attribute, now that it is, this logic is not required.

src/cosl/backends/loki.py
swetha1654: i think `backends` is a more appropriate name for the folder. Updated the code.

MichaelThamm:
To gain extra confidence that this PR is not breaking and providing the features we want, we should:
1. Create a tandem PR in OTLP charmlib, so we can review the API usage
2. @MichaelThamm reference this cosl PR from branch in Prometheus and run `tox` to test that there are no breaking changes.

swetha1654:
@MichaelThamm I have created a tandem [PR](https://github.com/canonical/charmlibs/pull/437) in charmlib to update the `otlp` interface to use the `GenericRules` class.

swetha1654:
> @MichaelThamm reference this cosl PR from branch in Prometheus and run tox to test that there are no breaking changes.

I think the linting will most likely fail. You would need to add `# pyright: ignore[reportPrivateImportUsage]` to wherever `HOST_METRICS_MISSING_RULE_NAME` and `generic_alert_groups` is imported. Or you would have to change the import line to import these from `cosl.common_rules` instead.

src/cosl/backends/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Rule backends for GenericRules."""
```
MichaelThamm:
```suggestion
"""Implementations of RuleBackend."""
```

src/cosl/backends/grouped_rules.py
```diff
@@ -0,0 +1,186 @@
...
+    OfficialRuleFileFormat,
+    OfficialRuleFileItem,
+    QueryType,
+    SingleRuleFormat,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class _GroupedRuleBackend(RuleBackend[OfficialRuleFileItem]):  # type: ignore
```
MichaelThamm, line 33: Reminder to check if this ignore is still needed.

src/cosl/rules.py
```diff
@@ -208,10 +245,232 @@ def __init__(
...
+                return []
+
+        try:
+            return self.from_dict(rule_file, **kwargs)
+        except ValueError as e:
+            logger.error("Invalid rules file: %s (%s)", file_path.name, e)
+            return []
+
+    @abstractmethod
+    def validate(self, rules: Dict[str, List[T]]) -> Tuple[bool, str]:
```
MichaelThamm, line 336: Shouldn't this return a `Result`?
```suggestion
    def validate(self, rules: Dict[str, List[T]]) -> Result:
```

src/cosl/rules.py
```diff
@@ -208,10 +245,232 @@ def __init__(
...
+            error message.
+        """
+        if not rules:
+            return Result(rules=self.backend.as_dict([]), errmsg=None)
+
+        output = dict(rules)
+        valid, errmsg = self.backend.validate(output)
+        if not valid:
+            return Result(rules=output, errmsg=errmsg)
+        return Result(rules=output, errmsg=None)
```
MichaelThamm, line 444: If the backend.validate also returned `Result` instead of a Tuple, then we can just return that instead:
```suggestion
        output = dict(rules)
        return self.backend.validate(output)
```

src/cosl/rules.py
```diff
@@ -73,13 +73,45 @@
 - `juju_model`
 - `juju_model_uuid`
 - `juju_application`
-"""  # noqa: W505
+
+## Generic Rules (Latest)
+
+The ``GenericRules`` class is a format-agnostic aggregator that collects alerting,
```
MichaelThamm: This has a slightly weird naming problem. We have documentation in multiple places and officially:
- https://documentation.ubuntu.com/observability/latest/explanation/alerting/generic-rules/

stating that the injected rules, not specific to any charm are called "generic rules". I noticed that you were trying to move the generic rules to `common_rules.py`. I think we should keep the name "generic rules" and rename this class to `AgnosticRules`, `RulesV2` or something else. WDYT?

src/cosl/backends/grouped_rules.py
MichaelThamm, line 1: It seems you removed a lot of the original docstrings. I think the existing one for rules.py still has some good context, which is now relevant only for the grouped_rules.py file. Can you please add some of it back?

I rewrote it slightly:

---

"""YOUR EXISTING DOCSTRING

This backend supports alerting and recording rule types, which can be added from a file, directory, or raw dict via ``add()`` and ``add_path()``. A charm's bundled rules can be added from file, from a directory conventionally named as one of:
- `prometheus_alert_rules`
- `prometheus_recording_rules`
- `loki_alert_rules`
- `loki_recording_rules`

This directory must reside at the top level in the `src` folder of the consumer
charm. Each file in this directory is assumed to be in one of two formats:
- the official Prometheus rule format, conforming to the
[Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/)
- a single rule format, which is a simplified subset of the official format,
comprising a single alert rule per file, using the same YAML fields.

The file name must have one of the following extensions:
- `.rule`
- `.rules`
- `.yml`
- `.yaml`

An example of the contents of such a file in the custom single rule
format is shown below.

```
alert: HighRequestLatency
expr: job:request_latency_seconds:mean5m{my_key=my_value} > 0.5
for: 10m
labels:
  severity: Medium
  type: HighLatency
annotations:
  summary: High request latency for {{ $labels.instance }}.
```

The specific Loki or Prometheus ``_GroupedRuleBackend`` implementation will read all available rules and
also inject "filtering labels" into the expressions. The
filtering labels ensure that rules are localised to the metrics
provider charm's Juju topology (application, model and its UUID). Such
a topology filter is essential to ensure that rules submitted by
one provider charm generates information only for that same charm. When
rules are embedded in a charm, and the charm is deployed as a
Juju application, the rules from that application have their
expressions automatically updated to filter for metrics/logs coming from
the units of that application alone. This removes risk of spurious
evaluation, e.g., when you have multiple deployments of the same charm
monitored by the same Prometheus or Loki.

Not all rules one may want to specify can be embedded in a
charm. Some rules will be specific to a user's use case. This is
the case, for example, of rules that are based on business
constraints, like expecting a certain amount of requests to a specific
API every five minutes. Such alerting or recording rules can be specified
via the [COS Config Charm](https://charmhub.io/cos-configuration-k8s),
which allows importing alert rules and other settings like dashboards
from a Git repository.

Rules generated will automatically include Juju topology labels. These labels indicate the source of the record or alert. The following labels are automatically included with each rule:

- `juju_model`
- `juju_model_uuid`
- `juju_application`
"""  # noqa: W505

---

@swetha1654 I did not format it correctly, but the content is approved.

MichaelThamm (changes requested):
I think this PR is pretty much ready to be merged content-wise after my recent comments are resolved. However, there is one thing which makes me nervous: we are lacking a lot of rules coverage for the new rules class. You added some tests: `test_backend` and `test_generic_rules`, but these provide little coverage compared to the existing set of tests we had for the `Rules` class.

One could argue that since this lib will only be used by the OTLP charmlib, we there is minimal risk. I think this is a good way to faze in this feature into the rules ecosystem without too much friction for our users. We will have to migrate the tests slowly to use the new rules classes to gain the same confidence as before. I don't think this is your responsibility, and given that you will be an early adopter of this feature, I think we can proceed with merging. I will discuss this briefly with @sed-i just to be safe.

src/cosl/backends/grouped_rules.py
swetha1654: Yes it is. I get the error `_GroupedRuleBackend" is not accessed` if i don't add it. Pylance expects private classes to be used within the same module. But `_GroupedRuleBackend` is not used in this module, its used by `prometheus.py` and `loki.py`. The reason i kept private is because i don't want the class to be accessed outside cos-lib. Its meant to be internal to cos-lib only.

src/cosl/rules.py
swetha1654: Yeah that makes sense. Let me update this.

src/cosl/rules.py
swetha1654: Nope this would be Tuple. The `GenericRules` class will take care of returning `Result` from the output of this method.
`CosTool.validate_alert_rules` returns a tuple and i decided to keep the backend validate method to return the same as that.

src/cosl/rules.py
swetha1654: RuleBackend.validate does not return a Result, it returns a tuple. If i change that to return Result instead, i would have to update the `_GroupedRuleBackend.validate` method's implementation to read the output from CosTool and convert that to `Result` before returning it. I would have to do the same with sigma backend when i implement it in the next PR. Since this is common logic to all backends i thought it would be better placed here. Let me know if you feel otherwise.

src/cosl/backends/grouped_rules.py
swetha1654: @MichaelThamm not sure i fully understand. This docstring is already there in `rules.py`, i haven't removed any of the original documentation. 

I have updated the `rules.py` docstring to include additional information about the new classes, keeping the existing docs intact.

swetha1654: > This backend supports alerting and recording rule types, which can be added from a file, directory, or raw dict via add() and add_path()

The backend implementation does not expose add or add_path. That is exposed by the `AgnosticRules` class, so the documentation makes sense to live there.

> The specific Loki or Prometheus _GroupedRuleBackend implementation will read all available rules and
also inject "filtering labels" into the expressions. 

This documentation is better placed at `loki.py` and `prometheus.py` modules since thats where the backends are imported from.

sed-i:
Thanks for this @swetha1654!
As you discovered, our impl of the rules has some flaws, and I see how the current approach sets up the stage for adding more rule types.

Before we go all in on `Generic[T]` and `abc`, I was wondering if we could explore a simpler approach, where we avoid generalizing via base classes.

As you outlined in the comparison table in #191, `*QL`-rules have little in common with sigma rules:


|                | Prometheus                                                                                                                                   | Loki                                                                                                                                                              | Tempo                                                                                                                               |
| -------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
| Alerting rule  | PromQL, mimir ruler, sent to alertmanager ([ref](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#alerting-rules)) | LogQL, evaluated by loki ruler, sent to alertmanager ([ref](https://grafana.com/docs/loki/latest/alert/))                                                         | TraceQL, via metrics-generator, experimental ([ref](https://grafana.com/docs/grafana/latest/alerting/examples/trace-based-alerts/)) |
| Recording rule | PromQL, mimir rules, "pushed to self" ([ref](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules))   | LogQL, evaluated by loki ruler, produces metric that is persisted to WAL, sent to mimir ([ref](https://grafana.com/docs/loki/latest/operations/recording-rules/)) | N/A                                                                                                                                 |

Looking at [`_otlp.py`](https://github.com/canonical/charmlibs/blob/01477c2c02c5986161f53d9c4b146a31fe06ddf2/interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py#L137), is there conceptually much more to do than the following?

```diff
-from cosl.rules import HOST_METRICS_MISSING_RULE_NAME, Rules, generic_alert_groups
+from cosl.rules import HOST_METRICS_MISSING_RULE_NAME, Rules, SigmaRules, generic_alert_groups
```

```diff
@dataclass
class RuleStore:
    topology: JujuTopology
    logql: Rules = field(init=False)
    promql: Rules = field(init=False)
+   sigma: SigmaRules = field(init=False)
```

```diff
class _RulesModel(BaseModel):

    logql: OfficialRuleFileFormat = Field(...)
    promql: OfficialRuleFileFormat = Field(...)
+   sigma: SigmaRuleFileFormat = Field(
+       description='...',
+       default_factory=SigmaRuleFileFormat,
+   )
```

Then, for sigma rules, instead of inheriting ready-made implementations from parent classes, we could extract logic into free function that could be reused by SigmaRules.

Wdyt?

MichaelThamm:
@sed-i I do think your suggestion is the quickest way to get Sigma rules into otelcol. The main updates are required in the [rules](https://github.com/canonical/charmlibs/blob/01477c2c02c5986161f53d9c4b146a31fe06ddf2/interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py#L465) method for topology injection (if desired), and the RuleStore object.

cos-lib abstracts a lot of the rules effort for PromQL and LogQL (e.g., injection, validation, adding, as_dict, etc.). As long as we still have Sigma implementation detail abstracted in cos-lib (like we do for promQL and logQL), then I think this is a great solution. Similar to how [promql are added to the rules store: self.promql.add(...)](https://github.com/canonical/charmlibs/blob/01477c2c02c5986161f53d9c4b146a31fe06ddf2/interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py#L112) using methods from a Sigma rules class in cos-lib, would be trivial.

swetha1654:
@sed-i @MichaelThamm If i understand correctly, what you're proposing is a seperate SigmaRules class in `cos-lib` instead of having an abstract class and then implementing it for Sigma / Prometheus / Loki. 

> As you outlined in the comparison table in https://github.com/canonical/cos-lib/issues/191, *QL-rules have little in common with sigma rules

I agree that Sigma rules themselves don't have a lot in common with *QL-rules, but there is still a lot in common between what we want to do with these rules. For eg:
1. we want to expose `add/add_path` APIs for users to add these rules, regardless of rule type.
2. We want to expose a method to validate these rules, regardless of their type.
3. We want a way for the users to export these rules to a dict to dump them into relation data, regardless of type.
4. We want each rule-type to specify what type of file extensions are accepted, so different backend implementations don't miss out specifying all allowed extensions.

Implementing a seperate SigmaRules class would mean repeating a lot of common logic between the user facing APIs, possibilities of divergence in the expected behaviour of `add/add_path` APIs, and many more issues. I don't see any added value in implementing SigmaRules as a separate class. If anything, it could make future maintenance of these *Rules classes more difficult in the future as we now have multiple classes to maintain.

Let me know if you have any further questions. I have visualized what I am trying to do with this PR in a [diagram](https://github.com/canonical/cos-lib/issues/191#issuecomment-4333126477) in the original issue for better understanding.

MichaelThamm:
Closing this PR in favor of:
- https://github.com/canonical/cos-lib/pull/197

PR #195, opened 2026-05-08
refactor: Rules class helper methods as free functions

MichaelThamm:
## Issue

We have some static methods in `Rules` that we want to move to free functions so that they can be reused in other rules-like classes e.g., SigmaRules.

### Checklist

## Context

These static methods were private so we have relative confidence that they were not used externally.

Ref to other PRs for context:
- https://github.com/canonical/cos-lib/pull/196
- https://github.com/canonical/charmlibs/pull/467

## Testing Instructions

Use these free functions in the Rules class which is thoroughly tested by the existing unit tests in cosl.


src/cosl/rules.py
```diff
@@ -221,6 +221,55 @@ class InjectResult:
     errmsg: Optional[str]
 
 
+# --- Free functions for file I/O, reusable across rule backends --- #
+
```
MichaelThamm:
```suggestion
```

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "cosl"
-version = "1.9.2"
+version = "1.10.0"
```
MichaelThamm, line 7: Make sure this is the correct version before merging

lucabello (approved):

PR #196, opened 2026-05-08
feat: Sigma rules

MichaelThamm:
## Issue

We need a Sigma rules feature in cos-lib to unblock these PRs:
- https://github.com/canonical/charmlibs/pull/466
- https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/300

### What
Introduces a `SigmaRules` utility and the supporting type definitions so downstream charm libraries (e.g. `charmlibs.interfaces.otlp`) can aggregate and forward [Sigma](https://sigmahq.io/docs/basics/rules.html) detection rules with Juju topology, the same way `AlertRules`/`RecordingRules` handle PromQL/LogQL.

### `SigmaRules` (`src/cosl/rules.py`)
- `add(rule_dict)` — accepts a single Sigma rule or a `{"rules": [...]}` collection; skips entries missing required fields.
- `add_path(dir_path, *, recursive=False)` — aggregates rule files (`.rule/.rules/.yml/.yaml`) from a file or directory.
- `as_dict()` — returns the collection in `SigmaRuleFileFormat`.
- **Topology injection** into each rule's `tags` as `namespace.value` (e.g. `juju_model.foo`):
  - existing namespaces are **not** overwritten (caller-set `juju_*` tags win),
  - tags are **sorted** for deterministic relation data (avoids spurious `relation-changed` churn),
  - re-injection is idempotent.

Unlike PromQL/LogQL, Sigma rules have no grouping/`expr` concept, so there is no expression rewriting — only `tags` enrichment.

### Types (`src/cosl/types.py`)
- New `TypedDict`s: `SigmaRuleLogSource`, `SigmaRuleFormat`, `SigmaRuleFileFormat`.

### Tests
- `tests/test_rules_sigma.py` (348 lines) covering add/add_path/as_dict, topology injection precedence, deterministic tag ordering, and idempotent re-injection.
- Fixture rule files under `tests/sigma_rules/` (single rules + a collection).

### Notes
- Additive and backward compatible — no changes to existing `AlertRules`/`RecordingRules` behavior.

### Checklist


src/cosl/rules.py
```diff
@@ -608,3 +610,81 @@ class RecordingRules(Rules):
...
+
+        Args:
+            topology: an optional ``JujuTopology`` instance used to annotate all rules.
+        """
+        self.topology = topology
+        self.rules: List[SigmaRuleFormat] = []
+
+    def _inject_topology(self, rule: SigmaRuleFormat) -> SigmaRuleFormat:
+        """Inject juju topology labels into a sigma rule (in-place)."""
+        if self.topology:
```
MichaelThamm: We should copy the rule to not side-effect

src/cosl/rules.py
```diff
@@ -608,3 +608,88 @@ class RecordingRules(Rules):
...
 
     pass
+
+
+class SigmaRules:
+    """Utility class for amalgamating Sigma rule files and injecting juju topology.
+
+    Unlike Prometheus/Loki rules, Sigma rules are independent (no grouping concept):
+    detection logic lives in a ``detection`` block instead of an ``expr`` field. A rule
+    may carry an ``id`` (UUID) for reference, but it is optional and not enforced here as
```
MichaelThamm, line 618: Make sure to be careful with UUID since it can cause many relation-changed events. Ask Luca for details.

src/cosl/loki_logger.py
MichaelThamm, line 1: Revert this change

src/cosl/rules.py
MichaelThamm, line 228: Make sure that this works with Sigma rules. I think this is all-encompasing, but Sigma should filter to only YAML.

tests/test_rules_sigma.py
```diff
@@ -0,0 +1,348 @@
...
+    assert "juju_model.upstream-model" in tags
+    assert "juju_model.testmodel" not in tags
+    # other namespaces are still injected
+    assert "juju_application.myapp" in tags
+
+
+def test_existing_juju_model_uuid_tag_not_overwritten(sigma):
+    sigma.add(_rule("Upstream UUID", tags=["juju_model_uuid.deadbeef"]))
+    tags = sigma.rules[0]["tags"]
+    assert "juju_model_uuid.deadbeef" in tags
```
MichaelThamm, line 156: update name deadbeef

PR #198, opened 2026-06-16
feat: log files ignored due to unrecognized suffix

lucabello:
## Summary

- When `_multi_suffix_glob` filters out files because they don't carry a recognized suffix, emit an `info`-level log message listing the ignored paths and the expected suffixes.
- This gives operators visibility into why their rule files are silently skipped, without changing any existing behaviour.

## Motivation

Raised in [cos-configuration-k8s-operator#129](https://github.com/canonical/cos-configuration-k8s-operator/issues/129) and originally proposed for `loki-k8s-operator` in [#528](https://github.com/canonical/loki-k8s-operator/pull/528). Since `loki_push_api` v1 now delegates rule loading to `cosl`, the fix belongs here.

## Test plan

- [x] Existing test suite passes (`tox`)
- [x] Add a unit test


MichaelThamm (approved):
Please manually verify as per the checkbox item in the PR before merging!

lucabello:
Even better: I'm adding a unit test!

PR #202, opened 2026-07-10
perf: memoize `_exec` to avoid redundant `cos-tool` subprocess calls

Abuelodelanada:
## Issue

`CosTool` shells out to the external `cos-tool` binary once per alert expression (via `inject_label_matchers`) and once per rule file (via `validate_alert_rules`).

Each call spawns a subprocess and blocks reading its output over a pipe, which costs ~75-80ms of pure process overhead regardless of how trivial the transformation is.

At scale this dominates the caller's runtime. Profiling a charm that aggregates alert rules from 300 related applications showed a single reconciliation spending **~94% of its wall-clock time (≈45s of 48s at just 10 clients)** blocked in `BufferedReader.read`, waiting for `cos-tool` subprocesses to return.

The root cause is redundant work: the **same expression can be processed multiple times within a single run** (e.g. staged, then combined, then forwarded to several backends), and identical bundled rules are re-processed for every related application. `cos-tool` is invoked with the exact same arguments over and over.

## Solution

Memoize `CosTool._exec`, the single place where the `cos-tool` subprocess is spawned.

`cos-tool transform`/`validate` are pure, deterministic functions of their argument list (tool path + flags + expression/rule file), so identical invocations can safely return a cached result instead of re-spawning the process.

Implementation details:

- A module-level worker `_cached_exec(cmd: Tuple[str, ...])` decorated with `functools.lru_cache(maxsize=2048)` performs the actual `subprocess.run`.
- `_exec` now delegates to it: `return _cached_exec(tuple(cmd))`.
- The worker is a **module-level function, not an instance method**, so `self` is excluded from the cache key. This matters because callers create a short-lived `CosTool` per ruleset; a per-instance cache would miss almost everything and would also keep instances alive. The shared cache lets hits work across instances and across processing stages.
- `maxsize` is **bounded (2048), not `None`**, so the cache cannot grow unbounded in long-lived processes.
- A public `clear_exec_cache()` is exposed for callers that want to invalidate the cache (e.g. charms that reconcile their entire state on every event).

Measured impact on the reference charm (10 clients): reconciliation dropped from **~48s to ~13s**, with cos-tool invocations reduced from 601 to 141 (77% avoided).

### Checklist

## Context

- `_exec` is the only method in `CosTool` that spawns a subprocess; both `inject_label_matchers` (`cos-tool transform`) and `validate_alert_rules` (`cos-tool validate`) funnel through it, so caching there covers both paths.
- The cache lives in process memory. In a Juju charm each hook is a fresh, short-lived Python process, so the cache does not (and need not) persist across events — the redundant work being eliminated happens **within a single reconciliation**, where the same expressions are reprocessed several times.
- Correctness caveat: the result depends on the `cos-tool` binary at the path embedded in the command. The binary is assumed immutable for the lifetime of the process. If that assumption is ever violated, call `clear_exec_cache()`. This is documented in the code.

## Testing Instructions

Run the cos-tool unit tests:

```bash
tox -e unit -- -k test_cos_tool.py 
```

New tests in `tests/test_cos_tool.py` (class `TestExecCache`) mock `subprocess.run` and assert:
- two identical invocations spawn the subprocess only once (second is a cache hit),
- distinct invocations each spawn a subprocess,
- the cache is shared across `CosTool` instances.

The existing `promql`/`logql` tests (which exercise the real binary) continue to pass, confirming the cache does not alter functional behavior.

## Upgrade Notes

No API changes and no behavioral changes for existing callers — `inject_label_matchers`, `validate_alert_rules` and `apply_label_matchers` return the same results as before.
The only additions are the internal memoization and the new public helper `cosl.cos_tool.clear_exec_cache()`. Bumped `project.version` from `1.10.0` to `1.10.1` release accordingly. No action is required from consumers to benefit from the speedup.

### DIFF
--- a/src/cosl/cos_tool.py
+++ b/src/cosl/cos_tool.py
@@ -33,9 +33,9 @@ _EXEC_CACHE_MAXSIZE = 2048
 def _cached_exec(cmd: Tuple[str, ...]) -> str:
     """Run a cos-tool command and memoize its output.
 
-    ``cos-tool transform``/``validate`` are pure, deterministic functions of their
+    `cos-tool transform` / `validate` are pure, deterministic functions of their
     argument list (tool path + flags + expression/rule file), so identical
-    invocations can be memoized. ``cmd`` is a tuple (hashable) rather than a list.
+    invocations can be memoized.
 
     Note:
         The result depends on the cos-tool binary at the path embedded in ``cmd``.
--- a/src/cosl/cos_tool.py
+++ b/src/cosl/cos_tool.py
@@ -8,6 +8,7 @@ import platform
 import re
 import subprocess
 import tempfile
+from functools import lru_cache
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -19,6 +20,42 @@ from .types import OfficialRuleFileFormat, QueryType
 logger = logging.getLogger(__name__)
 
 _F = TypeVar("_F", bound=Callable[..., Any])
+
+# Upper bound for the memoization cache of cos-tool invocations. cos-tool is invoked
+# once per alert expression, and the same expressions are frequently reprocessed
+# (e.g. bundled rules across many related applications, or the same rules staged then
+# forwarded to several backends). A bounded LRU cache avoids the (dominant) subprocess
+# spawn + pipe-read overhead while keeping memory usage bounded in long-lived processes.
+_EXEC_CACHE_MAXSIZE = 2048
+
+
+@lru_cache(maxsize=_EXEC_CACHE_MAXSIZE)
+def _cached_exec(cmd: Tuple[str, ...]) -> str:
+    """Run a cos-tool command and memoize its output.
+
+    ``cos-tool transform``/``validate`` are pure, deterministic functions of their
+    argument list (tool path + flags + expression/rule file), so identical
+    invocations can be memoized. ``cmd`` is a tuple (hashable) rather than a list.
+
+    Note:
+        The result depends on the cos-tool binary at the path embedded in ``cmd``.
+        The binary is assumed immutable for the lifetime of the process. Call
+        :func:`clear_exec_cache` if that assumption is ever violated.
+    """
+    result = subprocess.run(
+        list(cmd), check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    )
+    return result.stdout.decode("utf-8").strip()
+
+
+def clear_exec_cache() -> None:
+    """Clear the memoization cache of cos-tool invocations.
+
+    Charms that reconcile their whole state on every event may call this at the start
+    of a reconciliation to bound the cache to a single reconciliation, though the
+    bounded LRU size already prevents unbounded growth.
+    """
+    _cached_exec.cache_clear()
 
 
 def ensure_querytype(func: _F) -> _F:
@@ -192,5 +229,8 @@ class CosTool:
         return None
 
     def _exec(self, cmd: List[str]) -> str:
-        result = subprocess.run(cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
-        return result.stdout.decode("utf-8").strip()
+        # Delegate to a module-level, memoized worker. The result depends only on the
+        # command (not on instance state), so identical invocations across the many
+        # short-lived CosTool instances created during rule processing are deduplicated,
+        # avoiding redundant subprocess spawns.
+        return _cached_exec(tuple(cmd))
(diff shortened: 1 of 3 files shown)


src/cosl/cos_tool.py
```diff
@@ -20,6 +21,42 @@
...
+# Upper bound for the memoization cache of cos-tool invocations. cos-tool is invoked
+# once per alert expression, and the same expressions are frequently reprocessed
+# (e.g. bundled rules across many related applications, or the same rules staged then
+# forwarded to several backends). A bounded LRU cache avoids the (dominant) subprocess
+# spawn + pipe-read overhead while keeping memory usage bounded in long-lived processes.
+_EXEC_CACHE_MAXSIZE = 2048
+
+
+@lru_cache(maxsize=_EXEC_CACHE_MAXSIZE)
+def _cached_exec(cmd: Tuple[str, ...]) -> str:
```
MichaelThamm, line 33: Does this only hit the cache for the command string? If so, is there a way that the rule content could have changed, but the file name stays the same. E.g., injecting topology labels, which would then not receive a cos-tool invocation afterwards.

Abuelodelanada: This hit the cache for the command string and all its arguments, that's why it receives a tuple which is hashable.

For instance this function can receive a tuple like this one:

```python
("cos-tool", "--format", "promql", "transform", "--label-matcher=juju_model='my_model'", "--label-matcher=juju_application='app'", "-label-matcher=juju_unit='app/0'")
```

the output of the execution of this command is always the same.

sinapah (commented):
At a high-level this looks excellent Jose!

One minor note: it seems like in Prometheus Scrape and Prometheus Remote Write, we're still using COSTool based on a local implementation, not the one published by COS Lib. We should remember to fix those two libs for this to take effect.

Another small note: it looks like you made a choice to use the tuple of the CMD as the cache key. AFAICT, this should be fine and safe. However, wouldn't a hash like (expression, query_type, topology) be a more principled design? I think the PR is good as is, but it's just something to think about. In practice, it's unlikely for any factor to exist such that the CMD changes in a way that may cause a cache miss. Something to give a thought to, but as I said, this keying impl is fine by me as is!

Abuelodelanada:
> At a high-level this looks excellent Jose!
> 
> One minor note: it seems like in Prometheus Scrape and Prometheus Remote Write, we're still using COSTool based on a local implementation, not the one published by COS Lib. We should remember to fix those two libs for this to take effect.

Yes, this improvement should be implemented there as well!


 
> Another small note: it looks like you made a choice to use the tuple of the CMD as the cache key. AFAICT, this should be fine and safe. However, wouldn't a hash like (expression, query_type, topology) be a more principled design? I think the PR is good as is, but it's just something to think about. In practice, it's unlikely for any factor to exist such that the CMD changes in a way that may cause a cache miss. Something to give a thought to, but as I said, this keying impl is fine by me as is!

I'm using the tuple, but the lib internally uses the hash of that tuple, so there is no need to calculate a hash.

src/cosl/cos_tool.py
```diff
@@ -20,6 +21,42 @@
...
+    """Run a cos-tool command and memoize its output.
+
+    ``cos-tool transform``/``validate`` are pure, deterministic functions of their
+    argument list (tool path + flags + expression/rule file), so identical
+    invocations can be memoized. ``cmd`` is a tuple (hashable) rather than a list.
```
sed-i: Not sure why double backquote is needed.

```suggestion
    `cos-tool transform` / `validate` are pure, deterministic functions of their
    argument list (tool path + flags + expression/rule file), so identical
    invocations can be memoized.
```

tests/test_cos_tool.py
```diff
@@ -51,3 +52,87 @@ def test_typeerror_is_raised_if_no_query_is_used(self):
...
+    @unittest.mock.patch("cosl.cos_tool.subprocess.run")
+    def test_identical_commands_run_subprocess_once(self, mock_run):
+        """Identical cos-tool invocations should spawn the subprocess only once."""
+        mock_run.return_value = unittest.mock.Mock(stdout=b"transformed")
+        tool = CosTool(default_query_type="promql")
+
+        first = tool._exec(
+            [
+                "cos-tool",
+                "--format",
```
sed-i: Haven't looked at the details but odd that a CosTool instance would require us to indicate `cos-tool` and `--format`.
Also, not sure if this test currently gives us worthwhile lib coverage on top of lru_cache.

sed-i (commented):
Overall LGTM, a quick fix for painful problem!

MichaelThamm (approved):

PR #203, opened 2026-07-16
feat: persist the cos-tool result cache across processes

Abuelodelanada:
## Issue

`cos-tool` is invoked once per alert-rule expression (to inject Juju topology label matchers and to validate rules), and each invocation is a subprocess spawn costing on the order of tens of milliseconds. It is the dominant cost of rule processing in aggregator charms.

Two problems made this worse than necessary:

1. **The memoization cache did not survive across processes.** The previous cache was an in-process `functools.lru_cache`. Some charms may reconcile their whole state on every Juju event (for instance [`opentelemetry-collector-k8s`](https://github.com/canonical/opentelemetry-collector-k8s-operator/blob/b1a9814d6a18f7259ac32da2fbc6a9fe81bce235/src/charm.py#L244)), and each event is a **fresh process**, so the cache always started cold: every event re-ran `cos-tool` for every expression, even when nothing changed (e.g. `update-status`). At hundreds of related applications this pushes rule processing past Juju's hook timeout.

2. **Rule validation was never memoized.** `validate_alert_rules` wrote the rules to a randomly-named tempfile and passed that path to `cos-tool`. Because the path was part of the cache key and changed on every call, validation never hit the cache — not even within a single process.

## Solution

- **Persist the cache on disk, keyed by the (deterministic) inputs of each `cos-tool` invocation.** The in-process `lru_cache` is replaced by [`diskcache`], a small, pure-Python, SQLite-backed cache. Results computed in one process are reused by subsequent processes.
- **Add `configure_cache(cache_dir)`**, a module-level function that points the cache at a specific directory (e.g. a charm's persistent storage). By default the cache uses a fixed shared location (`/tmp/cosl-cos-tool`), so even callers that never configure it get cross-process reuse; `configure_cache` is for pointing at storage that survives a reboot / pod recreation. Passing `None` reverts to the default location.
- **Fix validation memoization.** `validate_alert_rules` now keys the cache on the rule *content* rather than on the nondeterministic tempfile path, so identical rules are validated by `cos-tool` only once.
- **Bound the cache by size** (`size_limit`, 256 MiB) with LRU eviction, so it never grows unbounded while staying hot for the expressions actually in use.

### Checklist

## Context

- **Every Juju event is a fresh Python process.** Any in-memory cache is therefore lost between events, which is why persistence across processes is the key win here — not just faster lookups within one process.
- **`cos-tool` output is deterministic.** For a given `(expression, topology, query_type)` the transformed expression (and the validation result) is always the same, so it is safe to memoize. The cache key is the exact argv passed to the binary (for `transform`), or a content-derived key (for `validate`).
- **`diskcache` uses SQLite and is process-safe**, though we do not rely on concurrent writers here: Juju serialises hook execution per unit.
- **Caching lives entirely inside `cosl`.** Consumers get persistence at a shared default location out of the box; they call `configure_cache(...)` only to point at storage that survives reboots / pod recreations.
- **The default path is a fixed `/tmp/cosl-cos-tool` (not a random temp dir).** A fixed, shared path lets all processes reuse the same cache without accumulating stray temp directories. It is safe without locking because Juju serialises hook execution per unit, so there is no concurrent writer. The hardcoded `/tmp` carries a `# noqa: S108` (acceptable in the single-user charm container); callers needing multi-user isolation should pass an explicit directory via `configure_cache`.

## Testing Instructions

```
tox -e unit
```

The unit test target downloads the real `cos-tool` binary if it is missing. New tests cover:

- Memoization: identical expressions run `cos-tool` once; distinct expressions run it per expression (exercised against the real binary).
- Persistence: a value written under a directory is reused after the cache is reopened (simulating a fresh process/event); the cache memoizes within a directory; reconfiguring to a different directory does not see the previous one's data; and the default location is the shared `/tmp/cosl-cos-tool`.
- Validation is memoized by content (previously it never hit the cache).
- **Fidelity:** the value served from the cache is byte-identical to what the real `cos-tool` binary returns — including after reopening a persisted cache in a fresh process. This guards the memoization layer introduced between the caller and the binary.

## Benchmarks

The attached [`cos_tool_cache_benchmark.py`](https://github.com/user-attachments/files/30074835/cos_tool_cache_benchmark.py) measures the impact of the cache against the real cos-tool binary. Run it from the repo root (with the binary present) as:

```shell
PYTHONPATH=src python3 benchmarks/cos_tool_cache_benchmark.py --counts 3000
```

`--counts N` is the number of distinct expressions to process (e.g. 3000 ≈ 300 related applications contributing ~10 alert expressions each). 
It reports four scenarios:

- `baseline` (one cos-tool subprocess per expression, no cache),
- `cold` (cache empty, every expression is a miss)
- `hot` (cache populated, all hits within the process)
- `cross-process hot` (a fresh process opens the populated on-disk cache — the realistic per-Juju-event case). 

A run with 3000 expressions:

```shell
$ PYTHONPATH=src python3 benchmarks/cos_tool_cache_benchmark.py --counts 3000
cos-tool cache benchmark
python: 3.12.3

==============================================================================
Expressions: 3000
==============================================================================

scenario                                       total (s)       ms/op   vs baseline
----------------------------------------------------------------------------------
baseline (subprocess, no cache)                 292.8440     97.6147    1.0x (ref)
cold (diskcache, all miss)                      292.8193     97.6064          1.0x
hot (diskcache, all hit)                          0.0747      0.0249       3919.4x
cross-process hot (fresh process, all hit)        0.0734      0.0245       3992.3x

```

The results confirm the design: a `cos-tool` invocation costs `~98` ms, so processing `3000` expressions uncached takes `~293` sec. The cache does not add measurable overhead on the cold path (cold ≈ baseline, since a miss still pays the subprocess plus a cheap SQLite write). Once populated, lookups are served from SQLite in ~0.025 ms — a `~3900×` speedup, dropping the total from `~293` sec to `~0.07` sec. Crucially, `cross-process hot` is as fast as `hot`, proving that opening the on-disk cache in a fresh process (what happens on every Juju event) is negligible: a charm that re-reconciles on `update-status` or any no-op event reprocesses all `3000` expressions in tens of milliseconds instead of minutes.

## Upgrade Notes

- Adds a new runtime dependency: **`diskcache`** (pure-Python, no native build, supports Python 3.8+).
- **No API break.** The cache now persists on disk by default at a shared location (`/tmp/cosl-cos-tool`) instead of living only in memory; behaviour within a process is otherwise the same (identical inputs are memoized). 
- Consumers that need caching to survive reboots / pod recreations should call `cosl.cos_tool.configure_cache(<persistent-dir>)` once at startup, pointing at storage that survives process restarts. For example, in a charm's `__init__`, pointing at a persistent storage mount declared in `charmcraft.yaml`:

  ```python
  from cosl.cos_tool import configure_cache

  class MyCharm(ops.CharmBase):
      def __init__(self, *args):
          super().__init__(*args)
          # `cache` is a filesystem storage mounted into the charm container.
          if storages := self.model.storages["cache"]:
              configure_cache(storages[0].location)
          ...
  ```

[`diskcache`]: https://grantjenks.com/docs/diskcache/

### DIFF
--- a/src/cosl/cos_tool.py
+++ b/src/cosl/cos_tool.py
@@ -8,11 +8,11 @@ import platform
 import re
 import subprocess
 import tempfile
-from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import yaml
+from diskcache import Cache  # pyright: ignore[reportMissingTypeStubs]
 from typing_extensions import TypeVar
 
 from .types import OfficialRuleFileFormat, QueryType
@@ -21,41 +21,76 @@ logger = logging.getLogger(__name__)
 
 _F = TypeVar("_F", bound=Callable[..., Any])
 
-# Upper bound for the memoization cache of cos-tool invocations. cos-tool is invoked
-# once per alert expression, and the same expressions are frequently reprocessed
-# (e.g. bundled rules across many related applications, or the same rules staged then
-# forwarded to several backends). A bounded LRU cache avoids the (dominant) subprocess
-# spawn + pipe-read overhead while keeping memory usage bounded in long-lived processes.
-_EXEC_CACHE_MAXSIZE = 2048
+# Upper bound (in bytes) for the on-disk cos-tool result cache. cos-tool is invoked once
+# per alert expression and its (deterministic) results are memoized to avoid the dominant
+# cost: the subprocess spawn (~tens of ms) on every reconcile. diskcache evicts least-
+# recently-used entries once this size is exceeded, so the cache never grows unbounded
+# while staying "hot" for the expressions actually in use. Entries are short strings, so
+# this comfortably holds the distinct expressions of a large (hundreds of apps)
+# aggregation deployment with room to grow.
+_EXEC_CACHE_SIZE_LIMIT = 256 * 1024 * 1024  # 256 MiB
+
+# Module-level, process-wide cache shared by all CosTool instances: cos-tool results
+# depend only on their inputs, not on which instance produced them, so results are
+# reusable across the many short-lived instances created during rule processing.
+#
+# Until configured with a persistent directory, it uses a temporary directory that lives
+# for the process (like the previous in-memory memoization). ``configure_cache`` points
+# it at persistent storage so results survive across process invocations (each Juju event
+# is a fresh process, which would otherwise always start cold).
+_exec_cache = Cache(size_limit=_EXEC_CACHE_SIZE_LIMIT)
 
 
-@lru_cache(maxsize=_EXEC_CACHE_MAXSIZE)
-def _cached_exec(cmd: Tuple[str, ...]) -> str:
-    """Run a cos-tool command and memoize its output.
+def configure_cache(cache_dir: Optional[Union[str, Path]]) -> None:
+    """Enable on-disk persistence of the cos-tool result cache.
 
-    `cos-tool transform` / `validate` are pure, deterministic functions of their
-    argument list (tool path + flags + expression/rule file), so identical
-    invocations can be memoized.
+    By default the cache lives in a temporary directory and starts cold in every process.
+    Charms that reconcile on every Juju event spawn a fresh process each time, so they
+    never benefit from a previous run's work. Point this at a persistent directory (e.g. a
+    charm's persistent storage mount) to carry the cache across process invocations.
 
-    Note:
-        The result depends on the cos-tool binary at the path embedded in ``cmd``.
-        The binary is assumed immutable for the lifetime of the process. Call
-        :func:`clear_exec_cache` if that assumption is ever violated.
+    Passing ``None`` reverts to a process-local temporary cache.
+
+    Args:
+        cache_dir: directory in which to store the cache, or ``None`` for a temporary one.
     """
-    result = subprocess.run(
-        list(cmd), check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-    )
-    return result.stdout.decode("utf-8").strip()
+    global _exec_cache
+    directory = str(cache_dir) if cache_dir is not None else None
+    _exec_cache.close()
+    _exec_cache = Cache(directory=directory, size_limit=_EXEC_CACHE_SIZE_LIMIT)
 
 
 def clear_exec_cache() -> None:
     """Clear the memoization cache of cos-tool invocations.
 
-    Charms that reconcile their whole state on every event may call this at the start
-    of a reconciliation to bound the cache to a single reconciliation, though the
-    bounded LRU size already prevents unbounded growth.
+    Charms that reconcile their whole state on every event may call this at the start of a
+    reconciliation to bound the cache to a single reconciliation, though the size limit
+    already prevents unbounded growth.
     """
-    _cached_exec.cache_clear()
+    _exec_cache.clear()
+
+
+def _exec(cmd: List[str], cache_key: Optional[Tuple[str, ...]] = None) -> str:
+    """Run a cos-tool command, memoizing its (deterministic) output.
+
+    Args:
+        cmd: the cos-tool command to run on a cache miss.
+        cache_key: a deterministic key identifying the invocation. Defaults to ``cmd``;
+            pass an explicit key for commands that reference a nondeterministic path
+            (e.g. the tempfile used by ``validate_alert_rules``), so the cache still hits.
+    """
+    key = tuple(cache_key) if cache_key is not None else tuple(cmd)
+    # diskcache ships no type stubs, so ``get``/``set`` are untyped; we only ever store
+    # ``str`` under these keys, so cast the retrieved value back to ``str``.
+    cached = cast(Optional[str], _exec_cache.get(key))  # pyright: ignore[reportUnknownMemberType]
+    if cached is not None:
+        return cached
+    result = subprocess.run(
+        list(cmd), check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    )
+    output = result.stdout.decode("utf-8").strip()
+    _exec_cache.set(key, output)  # pyright: ignore[reportUnknownMemberType]
+    return output
 
 
 def ensure_querytype(func: _F) -> _F:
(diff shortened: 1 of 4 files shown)


src/cosl/cos_tool.py
```diff
@@ -8,11 +8,11 @@
 import re
 import subprocess
 import tempfile
-from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import yaml
+from diskcache import Cache  # pyright: ignore[reportMissingTypeStubs]
```
sed-i, line 18: Should we try to use the stdlib instead?
For example something that relies on [shelve](https://docs.python.org/3/library/shelve.html).

Last commit to [diskcache](https://github.com/grantjenks/python-diskcache) is 2yr ago.

Abuelodelanada: My first implementation was using shelve, but then I decided to use diskcache, because with shelve the eviction / size bound logic was ~200 lines of code, whereas diskcache give us all that logic out-of-the-box.
For instance, with diskcache we can choose among [different eviction policies](https://grantjenks.com/docs/diskcache/tutorial.html#eviction-policies).

src/cosl/cos_tool.py
```diff
@@ -21,41 +21,85 @@
...
+_DEFAULT_CACHE_DIR = "/tmp/cosl-cos-tool"  # noqa: S108
+
+# Module-level, process-wide cache shared by all CosTool instances: cos-tool results
+# depend only on their inputs, not on which instance produced them, so results are
+# reusable across the many short-lived instances created during rule processing.
+#
+# It defaults to a fixed on-disk directory (see ``_DEFAULT_CACHE_DIR``). ``configure_cache``
+# points it at a different (e.g. charm-persistent) directory so results survive beyond what
+# ``/tmp`` offers (each Juju event is a fresh process, which would otherwise start cold).
+_exec_cache = Cache(directory=_DEFAULT_CACHE_DIR, size_limit=_EXEC_CACHE_SIZE_LIMIT)
```
MichaelThamm: Right now `_exec_cache = Cache(...)` runs at import, so importing `cosl.cos_tool` does
`os.makedirs("/tmp/cosl-cos-tool")` and can raise on a read-only/locked-down `/tmp`.
Recommend lazy initialization: hold only the directory at module scope and open the cache
on first use.

```python
_cache_dir: str = _DEFAULT_CACHE_DIR
_exec_cache: Optional[Cache] = None

def _get_cache() -> Cache:
    global _exec_cache
    if _exec_cache is None:
        _exec_cache = Cache(directory=_cache_dir, size_limit=_EXEC_CACHE_SIZE_LIMIT)
    return _exec_cache

def configure_cache(cache_dir: Optional[Union[str, Path]]) -> None:
    global _cache_dir, _exec_cache
    _cache_dir = str(cache_dir) if cache_dir is not None else _DEFAULT_CACHE_DIR
    if _exec_cache is not None:
        _exec_cache.close()
    _exec_cache = None  # reopened lazily at _cache_dir on next use

def clear_exec_cache() -> None:
    _get_cache().clear()
```

`_exec` / `get` / `set` then go through `_get_cache()`. Import becomes side-effect-free,
and `configure_cache` no longer eagerly creates directories it may never use.
(`test_default_cache_location` should assert on `_cache_dir` rather than
`_exec_cache.directory`.)

MichaelThamm: Can be a pain when:
- imported in a non-charm context (unit tests, a linter/doc tool, python -c "import cosl...") where creating /tmp/cosl-cos-tool is unwanted.
- A pre-existing /tmp/cosl-cos-tool causing `PermissionError` on open at import time.

src/cosl/cos_tool.py
```diff
@@ -21,41 +21,85 @@
...
-    result = subprocess.run(
-        list(cmd), check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-    )
-    return result.stdout.decode("utf-8").strip()
+    _exec_cache.clear()
 
 
-def clear_exec_cache() -> None:
-    """Clear the memoization cache of cos-tool invocations.
+def _exec(cmd: List[str], cache_key: Optional[Tuple[str, ...]] = None) -> str:
```
MichaelThamm, line 121: The `cos-tool` binary won't change between charm releases, and a
"clear the cache on refresh/upgrade" hook in the charm would cover it. But pushing that
onto every consumer (otelcol) is exactly the kind of defensive step that belongs in the library:
it's easy to forget, easy to get wrong, and the failure mode is *silent wrong output*.

Proposal: fold a cheap fingerprint into the cache key so any relevant change invalidates
entries automatically, no charm hook required. The fingerprint combines:

- the installed `cosl` version (`importlib.metadata.version("cosl")`), and
- a cheap stat of the binary (`st_size` + `st_mtime_ns`) — not a full hash, so it stays
  effectively free and is computed at most once per process.

```python
import functools
from importlib.metadata import version, PackageNotFoundError

@functools.lru_cache(maxsize=None)
def _cosl_version() -> str:
    try:
        return version("cosl")
    except PackageNotFoundError:
        return "0"

@functools.lru_cache(maxsize=None)
def _binary_fingerprint(path: str) -> str:
    # size + mtime is enough to detect a replaced binary and is far cheaper than hashing
    # a multi-MB executable on every invocation. Memoized per path for the process.
    try:
        st = os.stat(path)
        return f"{st.st_size}:{st.st_mtime_ns}"
    except OSError:
        return "missing"

def _fingerprint(binary_path: str) -> str:
    return f"cosl={_cosl_version()};bin={_binary_fingerprint(binary_path)}"
```

Then prefix every key (both `transform` and `validate`) with `_fingerprint(str(self.path))`.
This makes upgrade-safety a property of the library, so:

- a new `cosl` release invalidates old entries automatically;
- a replaced `cos-tool` binary (even at the same path) invalidates automatically;
- charms don't need an upgrade-hook workaround, and forgetting one can't produce stale
  rules.

The charm-side "clear on upgrade" can stay to have a cache clearing path, but it stops being
load-bearing.

src/cosl/cos_tool.py
MichaelThamm: `Cache(directory=..., size_limit=...)` uses diskcache's default
`eviction_policy="least-recently-stored"`, so the "stays hot for the expressions
actually in use" claim doesn't hold. Either pass `eviction_policy="least-recently-used"`
or update the comment/PR text to say least-recently-*stored*.

src/cosl/cos_tool.py
```diff
@@ -21,41 +21,85 @@
...
+    _exec_cache = Cache(directory=directory, size_limit=_EXEC_CACHE_SIZE_LIMIT)
 
-@lru_cache(maxsize=_EXEC_CACHE_MAXSIZE)
-def _cached_exec(cmd: Tuple[str, ...]) -> str:
-    """Run a cos-tool command and memoize its output.
 
-    `cos-tool transform` / `validate` are pure, deterministic functions of their
-    argument list (tool path + flags + expression/rule file), so identical
-    invocations can be memoized.
+def clear_exec_cache() -> None:
```
MichaelThamm: With the shared default, `clear_exec_cache()` wipes the cache for every
consumer/unit on the host, which both defeats the persistence win and is a behavior change
despite being listed under "No API break." Worth a docstring note, and ideally the
fingerprinting removes most of the motivation for consumers to call it at all.

This is only true if we have a cache dir per unit; not shared. Maybe we do want this behavior: `clear_exec_cache` clears for all units. What did you have in mind for otelcol?

MichaelThamm (approved):

MichaelThamm:
Just a general comment about the PR description and the comments in source code. They seem to be overly verbose which contributes to more comments than code in the long-term. Please make an effort to make the comments more concise, since LLMs love to over-comment.

src/cosl/cos_tool.py
Abuelodelanada: Nice catch!

src/cosl/cos_tool.py
Abuelodelanada: No that I think more abouot this, nobody is using this function because it is not released yet ;-) .... and there is no reason to delete the cache since cos-tool is deterministic. I'll remove the function

